PR #241, opened 2022-08-05
docs(echo): create tutorial

felangel:
## Status

**READY**

## Description

- docs(echo): create tutorial

## Type of Change

- [ ] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [X] 📝 Documentation
- [ ] 🗑️ Chore

PR #242, opened 2022-08-05
docs(hello world): update anchor text

scarletteliza:
to match the new precedent


felangel:
This was fixed as part of #241

PR #259, opened 2022-08-10
feat(dart_frog)!: enhanced provider

### LINKED ISSUE #196: [proposal] feat: provider's create param should allow asynchronous functions
## Description

As a developer, I would like to inject dependencies like so:
```dart
return handler.use(
    provider<SecretManagerApi>((context) async {
        final client = await clientViaApplicationDefaultCredentials(scopes: []);
        return SecretManagerApi(client);
    }),
);
```
However, this will result in a complier error as the `create` function passed to `provider` must be synchronous.

I propose we change the `create` function's definition from 
```dart
T Function(RequestContext context) create
``` 
to 
```dart
FutureOr<T> Function(RequestContext context) create
```
## Pros and Cons
The upside would be that you can write code like the example above. The downside is that it would required the `read<T>` function to return `FutureOr<T>`, which would be a breaking change and required you to `await` every call to `read`.

example:
```dart
Future<Response> onRequest(RequestContext context) async {
  final secretManagerApi = await context.read<SecretManagerApi>();
  return Response();
}
```
I argue that's not a major breaking change because 
1. `dart_frog` is still very young and experimental
2. I would be willing to bet 90% of all `Handler`s are `async` anyway (for reading from a database or calling an external API, for example). Thus, this would just mean adding a few more `await`s in your code.

## Other Workarounds
There are a few workarounds that we've tried that work today:

**1. Inject a `Future` that returns the dependency**
```dart
return handler.use(
    provider<Future<SecretManagerApi>>((context) async {
        final client = await clientViaApplicationDefaultCredentials(scopes: []);
        return SecretManagerApi(client);
    }),
);
```
and to read
```dart
Future<Response> onRequest(RequestContext context) async {
  final secretManagerApi = await context.read<Future<SecretManagerApi>>();
  return Response();
}
```

This works, but I think the syntax of reading a `Future` of a dependency is pretty clunky imo.

**2. A `readAsync` extension on `RequestContext`**
```dart
extension RequestContextAsync on RequestContext {
  Future<T> readAsync<T>() => read<Future<T>>();
}
```
to inject:
```dart
return handler.use(
    provider<Future<SecretManagerApi>>((context) async {
        final client = await clientViaApplicationDefaultCredentials(scopes: []);
        return SecretManagerApi(client);
    }),
);
```
and to read:
```dart
Future<Response> onRequest(RequestContext context) async {
  final secretManagerApi = await context.readAsync<SecretManagerApi>();
  return Response();
}
```

This also works, but imo it would be confusing for new developers to decide when to use `read` and when to use `readAsync`, and it wouldn't be in official documentation.

---

Finally, I have a branch that displays this functionality here: https://github.com/mtwichel/dart_frog/tree/feat/async-create-functions (not PR ready because of testing, but could become so fairly quickly I believe)

I'm curious what everyone thinks! Thanks for everything VGV 💙

felangel:
## Status

**READY**

## Description

- feat: add `lazy` and `cache` parameters on `provider` (closes #196)
  - `lazy` and `cache` default to `true`

## Type of Change

- [X] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/dart_frog/lib/src/provider.dart
+++ b/packages/dart_frog/lib/src/provider.dart
@@ -1,10 +1,87 @@
+import 'dart:async';
+
 import 'package:dart_frog/dart_frog.dart';
 
-/// Provide a value to the current handler by calling [create] lazily.
-Middleware provider<T extends Object>(
-  T Function(RequestContext context) create,
-) {
+final _cache = <Type, Object>{};
+
+/// Provide an object to the current handler by calling [create].
+///
+/// If [lazy] is `true`, [create] will be called only when performing
+/// a lookup via `context.read`.
+/// If [lazy] is `false`, [create] will be called immediately when
+/// an incoming request is received regardless of if a lookup is performed.
+///
+/// By default all providers are [lazy], meaning [create] is called on demand.
+///
+/// {@template provider_cache}
+/// If [cache] is `true`, [create] will only be called once and the result
+/// will be cached for the lifetime of the application.
+/// If [cache] is `false`, [create] will be called multiple times and
+/// the provided object will be recomputed.
+///
+/// By default providers cache, meaning the provided object won't be recomputed.
+/// {@endtemplate}
+///
+/// ```dart
+/// Handler middleware(Handler handler) {
+///  return handler.use(provider<String>(create: (_) => 'Hello World!'));
+/// }
+/// ```
+Middleware provider<T extends Object>({
+  required T Function(RequestContext context) create,
+  bool cache = true,
+  bool lazy = true,
+}) {
+  T _create(RequestContext context) {
+    return cache ? _memo<T>(() => create(context)) : create(context);
+  }
+
   return (handler) {
-    return (context) => handler(context.provide(() => create(context)));
+    return (context) {
+      if (lazy) return handler(context.provide(() => _create(context)));
+      final value = _create(context);
+      return handler(context.provide(() => value));
+    };
   };
+}
+
+/// Provide a `Future<T>` to the current handler by calling [create].
+/// [create] will be called when an incoming request is received.
+///
+/// {@macro provider_cache}
+/// ```dart
+/// Handler middleware(Handler handler) {
+///  return handler.use(
+///    futureProvider<String>(create: (_) async => 'Hello World!'),
+///  );
+/// }
+/// ```
+Middleware futureProvider<T extends Object>({
+  required Future<T> Function(RequestContext context) create,
+  bool cache = true,
+}) {
+  Future<T> _create(RequestContext context) async {
+    return cache ? _asyncMemo<T>(() => create(context)) : create(context);
+  }
+
+  return (handler) {
+    return (context) async {
+      final value = await _create(context);
+      return handler(context.provide<T>(() => value));
+    };
+  };
+}
+
+T _memo<T extends Object>(T Function() create) {
+  final cachedValue = _cache[T] as T?;
+  if (cachedValue != null) return cachedValue;
+  final value = create();
+  return _cache[T] = value;
+}
+
+Future<T> _asyncMemo<T extends Object>(Future<T> Function() create) async {
+  final cachedValue = _cache[T] as T?;
+  if (cachedValue != null) return cachedValue;
+  final value = await create();
+  return _cache[T] = value;
 }
--- a/packages/dart_frog/lib/src/provider.dart
+++ b/packages/dart_frog/lib/src/provider.dart
@@ -4,7 +4,7 @@ import 'package:dart_frog/dart_frog.dart';
 
 final _cache = <Type, Object>{};
 
-/// Removes the cached value for all providers.
+/// Removes the cached values for all providers.
 void resetProviders() => _cache.clear();
 
 /// Removes the cached value for the provided value of type [T].
--- a/packages/dart_frog/lib/src/provider.dart
+++ b/packages/dart_frog/lib/src/provider.dart
@@ -51,7 +51,7 @@ Middleware provider<T extends Object>(
   };
 }
 
-/// Provide a `Future<T>` to the current handler by calling [create].
+/// Provides a `Future<T>` to the current handler by calling [create].
 /// [create] will be called when an incoming request is received.
 ///
 /// {@macro provider_cache}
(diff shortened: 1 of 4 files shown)


packages/dart_frog/lib/src/provider.dart
```diff
@@ -1,10 +1,90 @@
+import 'dart:async';
+
 import 'package:dart_frog/dart_frog.dart';
 
-/// Provide a value to the current handler by calling [create] lazily.
+final _cache = <Type, Object>{};
+
+/// Clears the in-memory provider cache.
+void clearProviderCache() => _cache.clear();
```
wolfenrain, line 8: Is there a reason this isn't exposed? Do we want to allow people to clear caches?

If so, it should have a generic type

wolfenrain (approved):

packages/dart_frog/lib/src/provider.dart
felangel: Is there a use case you have in mind?

wolfenrain: Clearing.. a cache of a provider 😅.

Let's say your API is a game and you the API level up through experience so it counts up until you hit a threshold, and then u reset the experience to 0 and show a "level up" message.

Bit of a convoluted example, but I dont see any direct downsides of exposing a way to clear a cache.

packages/dart_frog/lib/src/provider.dart
```diff
@@ -1,10 +1,90 @@
...
+  Future<T> Function(RequestContext context) create, {
+  bool cache = true,
+}) {
+  Future<T> _create(RequestContext context) async {
+    return cache ? _asyncMemo<T>(() => create(context)) : create(context);
+  }
+
+  return (handler) {
+    return (context) async {
+      final value = await _create(context);
```
mtwichel, line 72: If I understand correctly, does this means the objects will be created before the hander is called? If so, this doesn't really work for my particular use case since we have a lot of routes that don't need certain expensive dependencies and I'd rather not build them if we can avoid it (although the caching will help a lot with that).

I suppose we could just rework our routes so that the middleware is only provided when needed, but then we'd have to design our route structure based on functionality, not what makes logical sense (following REST for example).

If there was an optional parameter that was like `bool Function(RequestContext context) createWhen` so I could tell the provider "_only create this expensive dependency if a certain header is set_" , I think that would work well too. It may be a bit convoluted for beginners, but it'd be highly optional 🤷

mtwichel:
Overall I think this is a great improvement! 💙 I love the caching and the lazy loading. Great stuff!

One suggestion that may be out of scope for this PR (but maybe not 😉): could we add the concept of an optional key that allows you to provide, cache, and read multiple dependencies of the same type, and they'd be indexed by a program defined key.

My use case: we need to build an API client for Square APIs for some of our routes. This is an expensive operation because it involves reading from the database and decrypting the values, and both of those require reading from Google Secret manager. In total, 5 network calls to build the client (which is why I'm so excited about caching lol).

The problem is each merchant that we serve needs a separate API client (because the client has the decrypted access token in its memory and each merchant has a different token). So, it would be cool if I could provide a `SquareApiClient` for each merchant and index them by the merchant's ID.

The workaround I'm currently using is providing a wrapper object that manages the cache itself. I then provide the one wrapper object, then extract the `SquareApiClient` from the wrapper in the handler. This works fine, but I'd rather have all my caching done in one place and not too. 

The API I'm thinking of would look like this:

Providing
```dart
  return handler.use(futureProvider<SquareApiClient>(
    (context) => buildApiClient(context.request.headers['merchant-id']),
    customKey: (context) => context.request.headers['merchant-id'],
  );
}
```

Reading
```dart
Response onRequest(RequestContext context) {
  final value = context.read<SquareApiClient>(customKey: context.request.headers['merchant-id']);
  return Response(body: value);
}
```

I realize this is a pretty big feature request, but a man can dream 😆

packages/dart_frog/lib/src/provider.dart
felangel: Seems reasonable. Just to make sure I understand correctly, you're suggesting to expose something like:

```dart
/// Removes the cached value for the provided value of type [T].
void resetProvider<T>();
```

Do you feel it's still necessary to include an API to clear the entire cache?

packages/dart_frog/lib/src/provider.dart
felangel: > If I understand correctly, does this means the objects will be created before the hander is called?

In that case you could use a regular, lazy, cached provider and provide/read the `Future`. That would result in the value being computed only one time and only when it is read:

```dart
  return handler.use(provider<Future<AsyncValue>>((_) => _createAsyncValue());
}
```

Then you can access it via:

```dart
Future<Response> onRequest(RequestContext context) async {
  final asyncValue = await context.read<Future<AsyncValue>>();
} 
```

Hope that helps!

packages/dart_frog/lib/src/provider.dart
wolfenrain: That would solve my use case!

packages/dart_frog/lib/src/provider.dart
mtwichel: Would that cache the actual value or the `Future` of that value (since in that case it'd be using `_memo` and not `_asyncMemo`)? If it is caching the `Future`, would it recompute each time it's accessed?

felangel: It should cache the `Future` so the computation would only done once.

packages/dart_frog/lib/src/provider.dart
```diff
@@ -1,10 +1,93 @@
...
-/// Provide a value to the current handler by calling [create] lazily.
+final _cache = <Type, Object>{};
+
+/// Removes the cached values for all providers.
+void resetProviders() => _cache.clear();
+
+/// Removes the cached value for the provided value of type [T].
+void resetProvider<T>() => _cache.remove(T);
+
+/// Provide an object to the current handler by calling [create].
```
erickzanardo, line 13:
```suggestion
/// Provides an object to the current handler by calling [create].
```

erickzanardo (approved):
L🐸TM



Rogue Frog (1)

renancaraujo (approved):
lgvtm

felangel:
Closing this for now. After many discussions we feel these changes aren't necessary at this time and the existing `provider` API is flexible enough to be extended to support various use cases. We plan to include documentation for how to handle async providers as well as cache values shortly.

PR #370, opened 2022-10-13
docs(testing): include middleware tests

felangel:
## Status

**READY**

## Description

- docs(testing): include middleware tests

## Type of Change

- [ ] 🧹 Code refactor
- [X] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/docs/docs/basics/testing.md
+++ b/docs/docs/basics/testing.md
@@ -101,7 +101,7 @@ void main() {
 
 ## Middleware 🍔
 
-Unit testing middleware is very similar to unit testing route handlers -- they are both just dart functions after all!
+Unit testing middleware is very similar to unit testing route handlers — they are both just dart functions after all!
 
 ### Basics
 
--- a/docs/docs/basics/testing.md
+++ b/docs/docs/basics/testing.md
@@ -8,7 +8,7 @@ In Dart Frog, we can effectively unit test our route handlers and middleware usi
 
 ## Route Handlers 🚏
 
-Testing route handlers is pretty straightforward and doesn't require any new concepts because at the end of the day a route handler is just a plain Dart function.
+Testing route handlers is pretty straightforward and doesn't require any new concepts because a route handler is just a plain Dart function.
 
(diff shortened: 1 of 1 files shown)


docs/docs/basics/testing.md
```diff
@@ -4,9 +4,27 @@ sidebar_position: 5
 
 # Testing 🧪
 
-In Dart Frog, we can unit test our route handlers and middleware effectively because they are plain functions.
+In Dart Frog, we can effectively unit test our route handlers and middleware using [`package:test`](https://pub.dev/packages/test) and [`package:mocktail`](https://pub.dev/packages/mocktail).
 
-For example, we can test our route handler above using `package:test`:
+## Route Handlers 🚏
+
+Testing route handlers is pretty straightforward and doesn't require any new concepts because at the end of the day a route handler is just a plain Dart function.
```
scarletteliza, line 11:
```suggestion
Testing route handlers is pretty straightforward and doesn't require any new concepts because a route handler is just a plain Dart function.
```

docs/docs/basics/testing.md
```diff
@@ -22,15 +40,119 @@ class _MockRequestContext extends Mock implements RequestContext {}
...
       expect(response.body(), completion(equals(greeting)));
     });
   });
 }
 ```
 
-In the above test, we're using `package:mocktail` to create a mock `RequestContext` and stub the return value when calling `context.read<String>()`. Then, all we need to do is call `onRequest` with the mocked context and we can assert that the response is what we expect. In this case, we're checking the statusCode and response body to ensure that the response is a 200 with the provided greeting.
+## Middleware 🍔
+
+Unit testing middleware is very similar to unit testing route handlers -- they are both just dart functions after all!
```
scarletteliza, line 104:
```suggestion
Unit testing middleware is very similar to unit testing route handlers — they are both just dart functions after all!
```
em dash!!!!

scarletteliza (approved):

PR #383, opened 2022-10-17
fix(dart_frog_gen)!: cascading middleware

### LINKED ISSUE #382: Middleware in parent folder
```
Package dart_frog_cli is currently active at version 0.2.0.
dart_frog --version
0.2.0
```

**Description**

I guess I could use the solution below in previous releases, where each middleware in whole structure directories are called. 
Now it seems only in the file folder (not parent) and root __routes/_middleware.dart__ 

**Steps To Reproduce**

Suppose I have that structure...

```
routes/
  - _middleware.dart (1)
  - guards/
    - _middleware.dart (2)
    - feature1/
      -  _middleware.dart (3)
      - index.dart
    - feature2/
      - index.dart
  - public/
    ...
```

**Expected Behavior**

I want check bearer token in some routes and then i've grouped the features in one directory "guards/" and created the middleware 2 to achieve that.
But it's not being called in any subdirectories.

- In feature1, the middlewares are called respectively: 1 + 3
- In feature2, only the 1.

Should I repeat the middleware to check token (2) in each subdirectory in features?
I know I can call "use" for that, but what bothers me is the obligation to create the __middleware.dart_ in each subdirectory
In this moment I have more than 10 in my project and it will grow as we create new features...

```dart
  return handler.use(bearerAuthMiddleware());
}
```

I've checked the [docs](https://dartfrog.vgv.dev/docs/basics/middleware) and it's in line with what is happening. 

> {...} There can only ever be one piece of middleware per route directory with routes/_middleware.dart being middleware that is executed for all inbound requests.

I cannot change the root middleware (1) because some routes wouldn't need the token.

What I imagined about middlewares calls:

- In feature1: 1 + 3
- In feature2: 1 + 2  (as it doesn't have its middleware like feature1)
- In public: 1

Any other approach to achieve that kind of middleware token check?

Thank you.

felangel:
## Status

**READY**

## Description

- fix(dart_frog_gen)!: support cascading middleware (closes #382)

## Type of Change

- [X] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/dart_frog_gen/lib/src/build_route_configuration.dart
+++ b/packages/dart_frog_gen/lib/src/build_route_configuration.dart
@@ -64,6 +64,7 @@ List<RouteDirectory> _getRouteDirectories({
   required void Function(MiddlewareFile route) onMiddleware,
   required void Function(String endpoint, RouteFile file) onEndpoint,
   required void Function(RouteFile route) onRogueRoute,
+  List<MiddlewareFile> cascadingMiddleware = const [],
 }) {
   final directories = <RouteDirectory>[];
   final entities = directory.listSync().sorted();
@@ -73,20 +74,25 @@ List<RouteDirectory> _getRouteDirectories({
       ? directorySegment
       : '/$directorySegment';
   // Only add nested middleware -- global middleware is added separately.
-  MiddlewareFile? middleware;
+  MiddlewareFile? currentMiddleware;
   if (directory.path != path.join(Directory.current.path, 'routes')) {
     final middlewareFile = File(path.join(directory.path, '_middleware.dart'));
     if (middlewareFile.existsSync()) {
       final middlewarePath = path
           .relative(middlewareFile.path, from: routesDirectory.path)
           .replaceAll(r'\', '/');
-      middleware = MiddlewareFile(
+      currentMiddleware = MiddlewareFile(
         name: middlewarePath.toAlias(),
         path: path.join('..', 'routes', middlewarePath).replaceAll(r'\', '/'),
       );
-      onMiddleware(middleware);
+      onMiddleware(currentMiddleware);
     }
   }
+
+  final middleware = [
+    ...cascadingMiddleware,
+    if (currentMiddleware != null) currentMiddleware,
+  ];
 
   final files = _getRouteFiles(
     directory: directory,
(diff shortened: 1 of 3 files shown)


packages/dart_frog_gen/lib/src/build_route_configuration.dart
```diff
@@ -73,47 +74,54 @@ List<RouteDirectory> _getRouteDirectories({
...
     onRoute: onRoute,
     onRogueRoute: onRogueRoute,
   );
 
-  final baseRoute = directoryPath.toRoute();
-  for (final file in files) {
-    var endpoint = (baseRoute + file.route.toRoute()).replaceAll('//', '/');
-    if (endpoint.endsWith('/')) {
-      endpoint = endpoint.substring(0, endpoint.length - 1);
+  if (files.isNotEmpty) {
```
felangel, line 104: optimization which allows us to avoid mounting routers for routes with no route handlers.

packages/dart_frog_gen/lib/src/build_route_configuration.dart
```diff
@@ -73,47 +74,54 @@ List<RouteDirectory> _getRouteDirectories({
...
         path: path.join('..', 'routes', middlewarePath).replaceAll(r'\', '/'),
       );
-      onMiddleware(middleware);
+      onMiddleware(currentMiddleware);
     }
   }
 
+  final middleware = [
+    ...cascadingMiddleware,
+    if (currentMiddleware != null) currentMiddleware,
```
felangel, line 94: cascade middleware

packages/dart_frog_gen/test/src/build_route_configuration_test.dart
```diff
@@ -498,42 +488,146 @@ void main() {
...
+              'route': '/<name>',
+              'file_params': ['name']
+            }
+          ],
+          'directory_params': ['user']
+        },
+        {
+          'name': r'_$user_$id',
+          'route': '/<user>/<id>',
+          'middleware': [
```
felangel, line 529: example of cascading middleware

wolfenrain (approved):

PR #393, opened 2022-10-19
feat: Support reuse of nested router

davidmartos96:
## Status

**PENDING DISCUSSION**

## Description

Hello! I've just noticed that supporting nested routers has been implemented, great news.

The implementation is very similar to one of my early iterations in https://github.com/dart-lang/shelf/pull/288 . But according to jonasfj, creating a Router per each request could be resource intensive. 

What I propose is a way to access the mounted params from within the RequestContext/Request so that a Router can be created before hand.

The implementation I did in this PR is the same solution as in the last iteration in the `shelf_router` PR, but there might be other solutions. I'm open to discuss it.  I'm used to work with shelf and shelf_router as is, not with the dart_frog code generation, so there might be a solution that could fit nicely with it.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


felangel:
Hey @davidmartos96 👋
Thanks so much for opening a PR and for taking the lead on helping support mounting dynamic routes! 💙

> But according to jonasfj, creating a Router per each request could be resource intensive.

Not sure I follow, can you provide a bit more context regarding where excess Routers are dynamically created? In the meantime, I’ll run some benchmarks to gather some more quantitative data.

Thanks again! 🙏

davidmartos96:
@felangel 
> Not sure I follow, can you provide a bit more context regarding where excess Routers are dynamically created?

So taking the following snippet from the tests as an example:
```dart
    Handler createUserHandler(String user) {
      final router = Router()
        ..get('/self', (RequestContext context) {
          return Response(body: "I'm $user");
        })
        ..get('/', (RequestContext context) {
          return Response(body: '$user root');
        });
      return router;
    }

    final app = Router()
      ..mount('/users/<user>', (RequestContext context, String user) {
        final handler = createUserHandler(user);
        return handler(context);
      })
      ..all('/<_|[^]*>', (RequestContext context) {
        return Response(body: 'catch-all-handler');
      });
```

Each request to a resource in `/users/<something>` will be creating a different router, that as a result each of the nested router entries (get, put, ...) will be running a quite complex regular expression. 
https://github.com/VeryGoodOpenSource/dart_frog/blob/4038b3ccaadf2e4faa9ac6672e32ee558c02108d/packages/dart_frog/lib/src/router.dart#L305
https://github.com/VeryGoodOpenSource/dart_frog/blob/4038b3ccaadf2e4faa9ac6672e32ee558c02108d/packages/dart_frog/lib/src/router.dart#L266

So where I see it could be problematic is a situation where you have a nested router (`/users/<user>/`) and the user route has a quite some resources defined (10-20). Then for each request to any of those resources, 10-20 regexps will be processed. 
But I haven't benchmarked this situation. If you have the benchmark code from above and the tool you used I would be happy to give it a try.

Another point jonasf makes in the other thread is the following:
> Moreover it's nice that all routes are evaluated early on, so that issues in patterns throw at the server startup time.

Having the creation of the router inside a handler prevents you to early-detect a possible error in a subroute right when you start the server.

davidmartos96:
@felangel Hello!
I had some time to run some benchmarks (https://github.com/davidmartos96/dart_frog_dynamic_router_benchmark)

From my tests on my machine, with this PR, a dynamic nested router server performs 44% better. 
The benchmark code adds 20 subroutes to the nested router, which is where the bottleneck could appear. https://github.com/VeryGoodOpenSource/dart_frog/pull/393#issuecomment-1285089370

davidmartos96:
@felangel Any update on this? It would be great if you could take a look, as this could be beneficial to `dart_frog` in terms of performance. +40% difference could be noticeable at big scale.

felangel:
> @felangel Any update on this? It would be great if you could take a look, as this could be beneficial to `dart_frog` in terms of performance. +40% difference could be noticeable at big scale.

Sorry for the delay! I’m planning to revisit this sometime this week. Thanks again!

davidmartos96:
Any updates? :sweat_smile:

alestiago:
Hi @davidmartos96 , sorry for the delayed reply. We've not forgotten about this Pull Request. I will try looking at it, but it might take slightly longer than expected since I need to gather some context first.

renancaraujo (approved):

Issue #406, opened 2022-11-01
feat: `dart_frog dev` should be able to `--write-service-info`


Luckey-Elijah:
**Description**

When executing `dart` with `--enable-vm-service` you the ability to write a the "VM service info" file that holds a reference to the VM service URI. It's the URI from the `dart_frog dev` command:

```
The Dart VM service is listening on http://127.0.0.1:8181/hbDoUv07zvg=/
```

The file contents are JSON.
```json
{"uri":"http://127.0.0.1:8181/hbDoUv07zvg=/"}
```

This file enables VS Code (and potentially Android Studio) users to connect the debugger automatically through a launch config.  I'd like to see an option for `dart_frog dev` to write this file for debuggers to attach to.

## Example 
If `dart_frog dev` wrote the VM info file to `.dart_frog/vm-service-info.json`, (and using a `hello_world` project, like the `examples/`)

#### `.vscode/tasks.json`
```json
{
  "version": "2.0.0",
  "tasks": [
    {
      "label": "run hello world example",
      "command": "dart_frog",
      "args": ["dev", "--write-service-info"],
      "problemMatcher": [],
      "isBackground": true,
    }
  ]
}
```
#### `.vscode/launch.json`
```jsonc
{
  "version": "0.2.0",
  "configurations": [
    {
      "name": "Launch Then Attach dart_frog to debugger",
      "request": "attach",
      "type": "dart",
      "preLaunchTask": "run hello world example",

      // path to the "vm-service-info.json"
      "vmServiceInfoFile": "${workspaceFolder}/.dart_frog/vm-service-info.json",
      "deleteServiceInfoFile": true
    }
  ]
}
```

**Requirements**

- [ ] add `--write-service-info` flag to `dart_frog dev`
- [ ] writes `vm-service-info.json` to `dart_frog`

**Additional Context**

VSCode extension demonstrate support
https://github.com/Dart-Code/Dart-Code/issues/2424

Related `dart_frog` Issues:
 - #286

wolfenrain:
Since this is a duplicate of #286 where we are going to implement this, we will be closing this.

But for anyone looking to quickly get it up and running for themselves the above will work! Thank you for looking into this <3

PR #407, opened 2022-11-02
adds `--write-service-info` to `dart_frog dev`

Luckey-Elijah:
## Status

READY/**IN DEVELOPMENT**/HOLD

## Description

Adds support for #406.

Not sure of the behavior on Windows or Linux.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


felangel:
@Luckey-Elijah thanks for the PR! We did a similar exploration a while back and while it does work we discussed this internally and feel this approach is not the best long-term solutio. Instead it would likely be a better user experience if there was a first party [VSCode Debugger Extension](https://code.visualstudio.com/api/extension-guides/debugger-extension) to support this use-case so that developers don't have to manage local configuration.

Are you okay if we close this PR for now while we investigate the feasibility of the alternative approach? If we decide to revisit this approach I'm happy to re-open this PR.

Luckey-Elijah:
@felangel I agree, it's probably not the best option long-term solution. I'm okay with closing this PR!

PR #470, opened 2023-01-10
feat(dart_frog): cache body for Request and Response

### LINKED ISSUE #468: feat: be able to read body more than once on Request and Response
**Description**

I'm trying to add body validation on middleware level for both Request and Response, but look like those can be read only once which make the server crash the second time you want to read.

**Requirements**

- [ ] being able to read request body multiple time
- [ ] being able to read response body multiple time

For the request one workaround would be to read it and then provide it frog dependancy injection but didn't manage to make it work for now.

For response I have no workaround as it will be read by shelf to send back the response anyway, is there a way to duplicate the response so I can read the body in a safe way?

felangel:
## Status

**READY**

## Description

- feat(dart_frog): cache body for `Request` and `Response` (closes #468)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/dart_frog/lib/src/request.dart
+++ b/packages/dart_frog/lib/src/request.dart
@@ -95,7 +95,7 @@ class Request {
 
   final shelf.Request _request;
 
-  Completer<List<int>>? _bodyBytesCompleter;
+  Completer<String>? _bodyCompleter;
 
   /// Connection information for the associated HTTP request.
   HttpConnectionInfo get connectionInfo {
@@ -117,25 +117,21 @@ class Request {
     return HttpMethod.values.firstWhere((m) => m.value == _request.method);
   }
 
-  Future<List<int>> _bytes() async {
-    if (_bodyBytesCompleter == null) {
-      _bodyBytesCompleter = Completer<List<int>>();
-      final bytes = await _request.read().fold<List<int>>(
-        <int>[],
-        (previous, element) => previous..addAll(element),
-      );
-      _bodyBytesCompleter!.complete(bytes);
-    }
-    return _bodyBytesCompleter!.future;
-  }
-
   /// Returns a [Stream] representing the body.
-  Stream<List<int>> bytes() async* {
-    yield await _bytes();
-  }
+  Stream<List<int>> bytes() => _request.read();
 
   /// Returns a [Future] containing the body as a [String].
-  Future<String> body() => utf8.decodeStream(bytes());
+  Future<String> body() async {
+    if (_bodyCompleter == null) {
+      _bodyCompleter = Completer<String>();
+      try {
+        _bodyCompleter!.complete(await _request.readAsString());
+      } catch (error, stackTrace) {
+        _bodyCompleter!.completeError(error, stackTrace);
+      }
+    }
+    return _bodyCompleter!.future;
+  }
 
   /// Returns a [Future] containing the form data as a [Map].
   Future<Map<String, String>> formData() {
@@ -145,7 +141,7 @@ class Request {
   /// Returns a [Future] containing the body text parsed as a json object.
   /// This object could be anything that can be represented by json
   /// e.g. a map, a list, a string, a number, a bool...
-  Future<dynamic> json() async => jsonDecode(await _request.readAsString());
+  Future<dynamic> json() async => jsonDecode(await body());
 
   /// Creates a new [Request] by copying existing values and applying specified
   /// changes.
--- a/packages/dart_frog/lib/src/request.dart
+++ b/packages/dart_frog/lib/src/request.dart
@@ -95,8 +95,6 @@ class Request {
 
   shelf.Request _request;
 
-  Completer<String>? _bodyCompleter;
-
   /// Connection information for the associated HTTP request.
   HttpConnectionInfo get connectionInfo {
     return _request.context['shelf.io.connection_info']! as HttpConnectionInfo;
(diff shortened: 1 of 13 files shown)


packages/dart_frog/lib/src/request.dart
```diff
@@ -115,11 +117,29 @@ class Request {
...
+
   /// Returns a [Stream] representing the body.
-  Stream<List<int>> bytes() => _request.read();
+  Stream<List<int>> bytes() async* {
+    yield await _bytes();
+  }
 
   /// Returns a [Future] containing the body as a [String].
-  Future<String> body() => _request.readAsString();
+  Future<String> body() => utf8.decodeStream(bytes());
```
jaumard, line 138: you probably want to use body for json method or it will not get the cache

packages/dart_frog/lib/src/response.dart
```diff
@@ -50,18 +50,38 @@ class Response {
...
+
   /// Returns a [Stream] representing the body.
-  Stream<List<int>> bytes() => _response.read();
+  Stream<List<int>> bytes() async* {
+    yield await _bytes();
+  }
 
   /// Returns a [Future] containing the body as a [String].
-  Future<String> body() => _response.readAsString();
+  Future<String> body() => utf8.decodeStream(bytes());
```
jaumard, line 80: same here json doesn't get the cache, not sure about parseFormData btw

wolfenrain (approved):

packages/dart_frog/lib/src/request.dart
```diff
@@ -115,11 +117,25 @@ class Request {
...
+        (previous, element) => previous..addAll(element),
+      );
+      _bodyBytesCompleter!.complete(bytes);
+    }
+    return _bodyBytesCompleter!.future;
+  }
+
   /// Returns a [Stream] representing the body.
-  Stream<List<int>> bytes() => _request.read();
+  Stream<List<int>> bytes() async* {
```
jaumard, line 133: that's the one I wouldn't apply any cache, because if you have 10Gb of data together it is unlikely you want it cached in memory. You'll pipe to filesystem and that's it.

packages/dart_frog/lib/src/response.dart
```diff
@@ -50,18 +50,34 @@ class Response {
...
+        <int>[],
+        (previous, element) => previous..addAll(element),
+      );
+      _bodyBytesCompleter!.complete(bytes);
+    }
+    return _bodyBytesCompleter!.future;
+  }
+
   /// Returns a [Stream] representing the body.
-  Stream<List<int>> bytes() => _response.read();
```
jaumard, line 61: that's the one I wouldn't apply any cache, because if you have 10Gb of data together it is unlikely you want it cached in memory. You'll pipe to filesystem and that's it.

wolfenrain: That is a good point, maybe we should expose a flag so people can decide if they want it cached or not, and by default make it false? So it wouldn't be a breaking change for anyone that is already doing file transfers?

packages/dart_frog/lib/src/request.dart
felangel: yup good catch will address your comments in a few mins.

packages/dart_frog/lib/src/response.dart
felangel: Just pushed some changes so that everything except bytes is cached. Lmk what you think and thanks again for the review!

wolfenrain (approved):

packages/dart_frog/lib/src/response.dart
```diff
@@ -61,7 +63,17 @@ class Response {
   Stream<List<int>> bytes() => _response.read();
 
   /// Returns a [Future] containing the body as a [String].
-  Future<String> body() => _response.readAsString();
+  Future<String> body() async {
+    if (_bodyCompleter == null) {
+      _bodyCompleter = Completer<String>();
+      try {
+        _bodyCompleter!.complete(await _response.readAsString());
```
jaumard, line 70: hum I wonder if this actually works, because here you read it anyway, so shelf can't read it too to send back the response right? So not sure that's the way to go for this

packages/dart_frog/test/src/response_test.dart
```diff
@@ -45,6 +56,28 @@ void main() {
...
+
+    test('body can be read multiple times (async)', () async {
+      final body = json.encode({'test': 'body'});
+      final response = Response(body: body);
+
+      await expectLater(response.body(), completion(equals(body)));
+      await expectLater(response.body(), completion(equals(body)));
+
+      await expectLater(response.json(), completion(equals(json.decode(body))));
+      await expectLater(response.json(), completion(equals(json.decode(body))));
```
jaumard, line 78: this don't test that shelf if able to read the response and send it back, my guess is that will fail.

packages/dart_frog/lib/src/response.dart
jaumard: My go to for response would be to not create the shelf response until shelf need it to answer the request. but no idea if that would work that's just an idea

packages/dart_frog/test/src/response_test.dart
felangel: Why not? Shelf just creates the request and response object just like we are doing here. I verified e2e that the changes do work. Feel free to try it yourself and let me know if you encounter any issues, thanks!

packages/dart_frog/lib/src/response.dart
felangel: added some more tests around this, let me know what you think 👍

jaumard: Now that the response is cloned I guess it should work, I have no time to test it (maybe this weekend) but to me should be good, if it doesn't I'll open a new issue, so this don't stay open to long. Thanks for this!

jaumard (approved):
LGTM

renancaraujo (approved):

PR #668, opened 2023-05-29
fix: making non-dynamic routes take precedence over dynamic ones

### LINKED ISSUE #631: feat: Mount non-dynamic routes first
**Description**

Currently dynamic directory routes get mounted first which leads to other routes in the same directory never getting matched.

For example, when you have routes like `routes/api/[id]/index.dart` and `routes/api/status.dart` the latter can never get matched. The generated code will always mount the first route first so the router will always pass `status` to `[id]`.

**Additional Context**

Here is the root handler generated by dart_frog with my current project which implements the [Mastodon API](https://docs.joinmastodon.org/methods/accounts/#statuses). Routes for the second mount such as`/api/v1/accounts/search` can never be reached.

```dart
Handler buildRootHandler() {
  final pipeline = const Pipeline().addMiddleware(middleware.middleware);
  final router = Router()
    ..mount('/api/v1/accounts/<id>', (context,id,) => buildApiV1Accounts$idHandler(id,)(context))
    ..mount('/api/v1/accounts', (context) => buildApiV1AccountsHandler()(context))
    ..mount('/api/v1/timelines', (context) => buildApiV1TimelinesHandler()(context))
    ..mount('/api/v1', (context) => buildApiV1Handler()(context))
    ..mount('/oauth', (context) => buildOauthHandler()(context))
    ..mount('/', (context) => buildHandler()(context));
  return pipeline.addHandler(router);
}
```

### LINKED ISSUE #658: fix: dynamic routes override non dynamic ones.
**Description**

When declaring a dynamic route, they will always take precedence over a non dynamic on. For example, imagine that I have the following route:

`GET /turtles/[id]`

And the following:

`POST /turtles/random`

When calling `curl -X POST http://localhost:8080/turtles/random`, `routes/turtles/[id]/index.dart` will be called instead of `routes/turtles/random.dart`, which can be confusing.

**Steps To Reproduce**

1. clone https://github.com/erickzanardo/ninja_turtles_api
2. start the server
3. `curl -X POST http://localhost:8080/turtles/random`
4. See the 405 error

**Expected Behavior**

I understand that this can be considered an expected behaviour, one could argue that `random` is simply an `id` that the `index.dart` inside the /turtles/[id]/` folder should consider and handle. Which I completely agree, but on the other hand, `random.dart` and `[id]/index.dart` files are conflicting routes.

I believe there are two possible ways of fixing this:
 - Non dynamic routes should take precedence over dynamic ones, so we could check for non dynamic route handlers before checking in dynamics ones, that way the example above would work. I believe this to be a preferred way of handling this, being able to split handlers like this can lead to a more organized route structure where we would be able to better split similar routes in different files.
 - Identify these conflicts and give a warning to the user

erickzanardo:
## Status

**READY**

## Description

Fixes #658
Fixes #631 

## Type of Change

- [x] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/bricks/dart_frog_dev_server/hooks/pre_gen.dart
+++ b/bricks/dart_frog_dev_server/hooks/pre_gen.dart
@@ -72,11 +72,8 @@ Future<void> preGen(
 
   context.vars = {
     'port': context.vars['port'] ?? '8080',
-    'directories': configuration.directories
-        .map((c) => c.toJson())
-        .toList()
-        .reversed
-        .toList(),
+    'directories':
+        configuration.orderedDirectories().map((c) => c.toJson()).toList(),
     'routes': configuration.routes.map((r) => r.toJson()).toList(),
     'middleware': configuration.middleware.map((m) => m.toJson()).toList(),
     'globalMiddleware': configuration.globalMiddleware != null
--- a/bricks/dart_frog_prod_server/hooks/pre_gen.dart
+++ b/bricks/dart_frog_prod_server/hooks/pre_gen.dart
@@ -79,11 +79,8 @@ Future<void> preGen(
   );
 
   context.vars = {
-    'directories': configuration.directories
-        .map((c) => c.toJson())
-        .toList()
-        .reversed
-        .toList(),
+    'directories':
+        configuration.orderedDirectories().map((c) => c.toJson()).toList(),
     'routes': configuration.routes.map((r) => r.toJson()).toList(),
     'middleware': configuration.middleware.map((m) => m.toJson()).toList(),
     'globalMiddleware': configuration.globalMiddleware != null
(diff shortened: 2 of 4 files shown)


bricks/dart_frog_dev_server/hooks/pre_gen.dart
```diff
@@ -72,11 +72,8 @@ Future<void> preGen(
 
   context.vars = {
     'port': context.vars['port'] ?? '8080',
-    'directories': configuration.directories
-        .map((c) => c.toJson())
-        .toList()
-        .reversed
```
erickzanardo, line 78: @felangel do you remember why this `reversed` call was in here?

If I kept it, it would "nullify" the ordering set to make that non dynamic routes takes precedence.

I tested without the `reversed` call and it seems to be working fine, but I believe that this call must have had a reason to be in here before, maybe you remember and could advise me which cases it covers so I can test with the new changes?

felangel: iirc the idea was to order the directories from most nested to least nested (most specific to least specific) so that the router generation would mirror that order and routes like:

`/users/[id]/pets`

Would be registered and handled before routes like:

`/users/[id]`

erickzanardo:
Closing in favor of #687

PR #599, opened 2023-04-11
feat(dart_frog_cli): Add dart-vm-port option to dev command

easazade:
This will add a new option to `dart_fro_cli` `dev` command, named `--dart-vm-port` with the abbreviation `-d`, that enables defining a custom port for the dart VM service to listen on. 

So if this flag is used like this
`dart_frog -d 9191`
 the dart command that is going actually to run will be like this:
`dart --enable-vm-service=9191 .dart_frog/server.dart`

Running both [jaspr](https://github.com/schultek/jaspr/) and dart_frog in debug mode was not possible since dart VM was trying to use the same default port (8181) to listen on

## Status

**HOLD**
Should have opened an issue first. But If this change is something you can accept, please let me know to add some doc for it. Also tests.

## Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

## Preview

https://user-images.githubusercontent.com/38987302/231396262-9db4efc0-027a-4984-b2e1-d624cc91c8ac.mov

PR #669, opened 2023-05-29
feat: adding list command

erickzanardo:
## Status

**READY**

## Description

Adds a new list command to Dart frog, which will run the dart gen logic in order to gather all the endpoints available in the project, and print them in the console, making it easier for users get a summary of all the available endpoints in their project, without have to open the sourcebase in a file explorer.

This command can be improved in the future to show the HTTP method when we have a way of identifying them.

Screenshot 2023-05-29 at 17 12 57

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/dart_frog_cli/lib/src/commands/list/list.dart
+++ b/packages/dart_frog_cli/lib/src/commands/list/list.dart
@@ -1,0 +1,75 @@
+import 'dart:io';
+
+import 'package:args/command_runner.dart';
+import 'package:dart_frog_cli/src/command.dart';
+import 'package:dart_frog_gen/dart_frog_gen.dart';
+import 'package:mason/mason.dart';
+
+/// Definition for a function that builds a [RouteConfiguration] from a
+/// [Directory].
+typedef RouteConfigurationBuilder = RouteConfiguration Function(
+  Directory directory,
+);
+
+/// {@template list_command}
+/// `dart_frog list "path/to/project"`
+///
+/// Lists the routes on the project.
+/// {@endtemplate}
+class ListCommand extends DartFrogCommand {
+  /// {@macro list_command}
+  ListCommand({
+    super.logger,
+    RouteConfigurationBuilder buildConfiguration = buildRouteConfiguration,
+  }) : _buildConfiguration = buildConfiguration;
+
+  final RouteConfigurationBuilder _buildConfiguration;
+
+  @override
+  String get description => 'Lists the routes on a Dart Frog project.';
+
+  @override
+  String get name => 'list';
+
+  @override
+  final String invocation = 'dart_frog list "path/to/project"';
+
+  @override
+  Future<int> run() async {
+    final projectDir = _projectDirectory;
+
+    final configuration = _buildConfiguration(projectDir);
+
+    logger
+      ..info('Route list 🐸:')
+      ..info('==============\n');
+
+    for (final endpoint in configuration.endpoints.keys) {
+      logger.info(endpoint);
+    }
+
+    return ExitCode.success.code;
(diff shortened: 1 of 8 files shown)


packages/dart_frog_cli/lib/src/commands/list/list.dart
```diff
@@ -0,0 +1,75 @@
...
+
+  @override
+  Future<int> run() async {
+    final projectDir = _projectDirectory;
+
+    final configuration = _buildConfiguration(projectDir);
+
+    logger
+      ..info('Route list 🐸:')
+      ..info('==============\n');
```
alestiago, line 45: Nit: Remove header to simplify grep operations?

```suggestion
```

Perhaps for pretty prints, we can introduce add a subcommand to pretty print routes in a tree structure or table (in another PR, if proposal is agreed by the team or community).

packages/dart_frog_cli/lib/src/commands/list/list.dart
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  Directory get _projectDirectory {
+    final rest = results.rest;
+    _validateProjectDirectoryArg(rest);
+    return Directory(rest.first);
+  }
+
+  void _validateProjectDirectoryArg(List<String> args) {
+    if (args.isEmpty) {
```
alestiago, line 61: Nit: What do you think of using the current directory when no directory is specified?

Similar to `flutter run`?

Maybe instead of using raw args we can use a subcommand to specify the directory?

alestiago:
I have a small question regarding this. Say we merge https://github.com/VeryGoodOpenSource/dart_frog/pull/599, and now support multi development server. Then we may like to add a command to list all running servers with their information (port, etc) and perhaps some analytics (such as running time, request served, response time, etc).  At this point I would imagine we do `dart_frog dev --list` to do so? Or do you think `dart_frog list --servers` and `dart_frog list --routes` are more appropriate? 

I think the `list` command as it is right now it is useful and intuitive. I'm just trying to bring attention to the fact that it might be ambiguous what it does as the ecosystem grows and there might be more things you can "list".

renancaraujo (approved):

packages/dart_frog_cli/lib/src/commands/list/list.dart
erickzanardo: I like the idea of having a grep friendly output. But I feel that this would mostly be meant by human eyes, so I would argue that this should be the default.

I am gonna go ahead and implement a flag for this "cleaner" output.

And for the additional info, such as total number, etc, we sure can do in a follow up PR :)

erickzanardo:
> I have a small question regarding this. Say we merge #599, and now support multi development server. Then we may like to add a command to list all running servers with their information (port, etc) and perhaps some analytics (such as running time, request served, response time, etc). At this point I would imagine we do `dart_frog dev --list` to do so? Or do you think `dart_frog list --servers` and `dart_frog list --routes` are more appropriate?
> 
> I think the `list` command as it is right now it is useful and intuitive. I'm just trying to bring attention to the fact that it might be ambiguous what it does as the ecosystem grows and there might be more things you can "list" (eg middleware, servers, etc). In other words, how to we foresee scaling this command?

I feel we should have a list command that is generic, and list everything, on this PR, I implementing listing only the routes. In the future, we can expand it to list middlewares as well, although I am not sure I see much usefulness on that.

And if we want to allow the user to list only one thing or another, we can make more specific commands or flag that will "filter" the output.

I believe that is enough to keep us future proof :)

PR #687, opened 2023-06-06
feat: detect conflict between dynamic routes and non dynamic ones

### LINKED ISSUE #658: fix: dynamic routes override non dynamic ones.

erickzanardo:
## Status

**READY**

## Description

This PR adds a validation to detect dynamic routes that would eventually "override" non dynamic ones and adds an error to notify the user, so that route issue don't go unnoticeable.

Ideally, we would try to order the routes in a way that the non dynamics ones would take precedence over the dynamic, but that proved a complicated implementation which would bring many other complications related to middlewares.

Since the user can still simply use a single dynamic route to both handle a `GET /pets/123` and a `GET /pets/mine`, we decided to go with marking routes as conflicted when we encounter such cases.

Example output:
Screenshot 2023-06-06 at 10 57 32

Fixes #658
Fixes: #631

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/dart_frog_gen/lib/src/validate_route_configuration/route_conflicts.dart
+++ b/packages/dart_frog_gen/lib/src/validate_route_configuration/route_conflicts.dart
@@ -20,8 +20,40 @@ void reportRouteConflicts(
   /// Callback called when any route conflict is found.
   void Function()? onViolationEnd,
 }) {
-  final conflictingEndpoints =
+  final directConflicts =
       configuration.endpoints.entries.where((entry) => entry.value.length > 1);
+
+  final indirectConflicts = configuration.endpoints.entries.where((entry) {
+    final match = configuration.endpoints.keys.where((other) {
+      final keyParts = entry.key.split('/');
+      if (other == entry.key) {
+        return false;
+      }
+
+      final otherParts = other.split('/');
+
+      var match = false;
+
+      if (keyParts.length == otherParts.length) {
+        for (var i = 0; i < keyParts.length; i++) {
+          if ((keyParts[i] == otherParts[i]) ||
+              (keyParts[i].startsWith('<') || otherParts[i].startsWith('<'))) {
+            match = true;
+          } else {
+            match = false;
+            break;
+          }
+        }
+      }
+
+      return match;
+    }).isNotEmpty;
+
+    return match;
+  });
+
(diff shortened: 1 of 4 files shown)


felangel:
Why do we have output two conflicts per route pair?

erickzanardo:
> Why do we have output two conflicts per route pair?

Since they conflict with each other. I couldn't think how to decide if we show `A -> B` or `B -> A`, so I thought we should just simply show both.

I am happy to remove the "duplication" if we can come with a good logic on how to decide on which of the order to show

felangel:
> @felangel wdyt if we gave preference to the dynamic route, since it is the one that will handle most routes? Given a message like this:
> 
> ```
> both /turtles/random and /turtles/<id> resolve to /turtles/<id>
> ```
> 
> ?

Yup I think that makes sense and is what I expected when I originally saw the error.

wolfenrain (approved):

erickzanardo:
@felangel here is the new output

Screenshot 2023-06-08 at 09 57 35

renancaraujo (approved):

PR #717, opened 2023-06-21
feat(vscode): includes extension testing setup

alestiago:
## Status

**READY**

## Description

Adds the initial dependencies and setup to run test on the VS Code extension.

Changes:
- Adds dependencies that are/will required to test the extension
- Adds a simple extension test to be able to run the setup

## Demonstration

https://github.com/VeryGoodOpenSource/dart_frog/assets/44524995/affa774b-ed8a-4464-a988-5eeccc2e478a

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


extensions/vscode/src/test/runTest.ts
```diff
@@ -0,0 +1,29 @@
...
+      extensionTestsPath,
+      launchArgs: ["--disable-extensions"],
+    });
+  } catch (err) {
+    console.error("❌ Failed to run tests");
+    process.exit(1);
+  }
+}
+
+main().catch((err) => console.error(err));
```
erickzanardo, line 29:
```suggestion
main();
```

From what I got on the code, you don't need that catch as it will never be called since the whole function is wrapped in a try/catch, and in the case of an error, the error is already logged and the process is exited right after.

extensions/vscode/src/extension.ts
```diff
@@ -4,8 +4,12 @@ import * as vscode from "vscode";
  * This method is called when the extension is activated.
  *
  * The extension is activated the very first time the command is executed.
- * @param context
+ *
+ * @param {vscode.ExtensionContext} context
+ * @returns {vscode.ExtensionContext}
```
wolfenrain, line 9: Correct me if I am wrong, but these are no longer needed if you are using typescript right?

We should define a return type for this method on the method definition itself

wolfenrain (approved):

extensions/vscode/src/extension.ts
alestiago: I've moved the type to the function declaration. Thanks for catching that! 

I've left both @param and @returns to comply with [TS Doc](https://tsdoc.org/). Including them does alter the output when hovering on the functions.

erickzanardo (approved):

PR #815, opened 2023-07-27
feat: adding dart frog test

erickzanardo:
## Status

**IN DEVELOPMENT**

## Description

This PR adds a proposal for a `dart_frog_test` package which will help the testing of dart frog request handlers by offering an api that makes it easier to mock the RequestContext and Request to the developer.

I am marking this PR as draft as this is still a proposal and I am looking for feedback

At the point of this draft, I think the proposed API is what we need, and if people agree documentation and any missing tests will be written. But if you identify any missing pieces that could be added, please let me know.

This is the output of tests executed from the example folder:

```
$ dart test -r expanded
00:00 +0: test/routes/dice_test.dart: POST /dice: responds with a 200 and the rolled number.
00:00 +1: test/routes/dice_test.dart: POST /dice: responds with a 200 and the rolled number.
00:00 +2: test/routes/dice_test.dart: DELETE /dice: responds with method not allowed.
00:00 +3: test/routes/dice_test.dart: GET /dice: responds with method not allowed.
00:00 +4: test/routes/dice_test.dart: HEAD /dice: responds with method not allowed.
00:00 +5: test/routes/dice_test.dart: OPTIONS /dice: responds with method not allowed.
00:00 +6: test/routes/dice_test.dart: PATCH /dice: responds with method not allowed.
00:00 +7: test/routes/dice_test.dart: PUT /dice: responds with method not allowed.
```

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/dart_frog_test/lib/src/dart_frog_test.dart
```diff
@@ -0,0 +1,53 @@
...
+/// A test that can be used to test a Dart Frog handler.
+@isTest
+void testDartFrog(
+  String message, {
+  required String url,
+  required FutureOr<Response> Function(RequestContext request) onRequest,
+  HttpMethod method = HttpMethod.get,
+  Map<String, String>? headers,
+  dynamic requestJson,
+  String? requestBody,
```
alestiago, line 21: I'm not sure about including parameters for the Request. Maybe, would it be better if we allow the user to provide a `request` and they can do whatever they want with it?

I'm not sure, I might need to give it a shower thought 🚿

packages/dart_frog_test/lib/src/dart_frog_test.dart
```diff
@@ -0,0 +1,53 @@
...
+import 'package:mocktail/mocktail.dart';
+import 'package:test/test.dart';
+
+class _MockRequestContext extends Mock implements RequestContext {}
+
+class _MockRequest extends Mock implements Request {}
+
+/// A test that can be used to test a Dart Frog handler.
+@isTest
+void testDartFrog(
```
alestiago, line 14: The name doesn't convince me, I feel Dart Frog is way to generic as the subject of the test since we are testing a specific endpoint.

Maybe `testRoute`, `testMiddleware` and so on?

packages/dart_frog_test/lib/src/dart_frog_test.dart
erickzanardo: My reasoning behind having the fields of the Request received by the method was based on my incorrect assumption that instantiating a `Request` class was hard, which by now I see that it is not the case, so yes, this suggestion makes total sense! I am gonna refactor it!

packages/dart_frog_test/lib/src/dart_frog_test.dart
erickzanardo: yeah, good call, maybe `testRouteHandler`?

packages/dart_frog_test/lib/src/dart_frog_test.dart
```diff
@@ -0,0 +1,106 @@
...
+import 'package:test/test.dart';
+
+class _MockRequestContext extends Mock implements RequestContext {}
+
+/// {@template test_request}
+/// A model used to create a dart frog [Request] in a test.
+/// {@endtemplate}
+class TestRequest {
+  /// {@macro test_request}
+  TestRequest({
```
erickzanardo, line 15: Note: Ideally this class wouldn't exist, but Dart frog's `Request` class isn't super API friendly for tests since you:

(a) have to provide the method in string format, which is way less prone to typos than using `HttpMethod.x`
(b) you have to provide an `Uri`, which can be cumbersome to create since you need to provide the full path in order to use `Uri.parse`

In `TestRequest` use use the http method enum instead, making the test type safer and ask only for the path from the user, using a placeholder base url to make the tester easier to write and read.

packages/dart_frog_test/example/README.md
```diff
@@ -0,0 +1,12 @@
+# example
```
scarletteliza, line 1:
```suggestion
# Example
```

packages/dart_frog_test/example/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+[![style: very good analysis][very_good_analysis_badge]][very_good_analysis_link]
+[![License: MIT][license_badge]][license_link]
+[![Powered by Dart Frog](https://img.shields.io/endpoint?url=https://tinyurl.com/dartfrog-badge)](https://dartfrog.vgv.dev)
+
+An example application built with dart_frog
```
scarletteliza, line 7:
```suggestion
An example application built with Dart Frog
```

packages/dart_frog_test/lib/src/dart_frog_test.dart
alestiago: For (a) users could use the other constructors we provide as Request.get and so on, right?

erickzanardo: > For (a) users could use the other constructors we provide as Request.get and so on, right?

yes! but the default constructor is still available to them, so they could use and run into a confusing error

scarletteliza (approved):

packages/dart_frog_test/README.md
```diff
@@ -0,0 +1,69 @@
...
+```yaml
+dependencies:
+  dart_frog_test:
+```
+
+Install it:
+
+```sh
+dart pub get
+```
```
felangel, line 24: Can we show some example tests in the `README`?

packages/dart_frog_test/example/pubspec.yaml
```diff
@@ -0,0 +1,17 @@
+name: example
+description: An new Dart Frog application
+version: 1.0.0+1
+publish_to: none
+
+environment:
+  sdk: ">=3.0.0 <4.0.0"
+
+dependencies:
+  dart_frog: ^0.3.0
```
felangel, line 10:
```suggestion
  dart_frog: ^1.0.0
```

packages/dart_frog_test/example/pubspec.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+dev_dependencies:
+  dart_frog_test:
+    path: ../
+  mocktail: ^0.3.0
```
felangel, line 15:
```suggestion
  mocktail: ^1.0.0
```

packages/dart_frog_test/example/test/routes/index_test.dart
```diff
@@ -0,0 +1,22 @@
+import 'dart:io';
+
+import 'package:dart_frog_test/dart_frog_test.dart';
+import 'package:test/test.dart';
+
+import '../../routes/index.dart' as route;
+
+void main() {
+  testRouteHandler(
```
felangel, line 9: I'm not sure whether this API is more readable then the default tests personally:

```dart
import 'dart:io';

import 'package:mocktail/mocktail.dart';
import 'package:test/test.dart';

import '../../routes/index.dart' as route;

class _MockRequestContext extends Mock implements RequestContext {}

void main() {
  group('GET /', () {
    test('responds with a 200 and "Welcome to Dart Frog!".', () {
      final context = _MockRequestContext();
      final response = route.onRequest(context);
      expect(response.statusCode, equals(HttpStatus.ok));
      expect(
        response.body(),
        completion(equals('Welcome to Dart Frog!')),
      );
    });
  });
}
```

Imo the original test is much easier to follow/understand 🤷 

I'm not sure that a test abstraction is necessary

erickzanardo: The snippet you provided is missing the mocking of the Request right?

That is basically the only help that the abstraction provide

felangel: The snippet includes the mocking of the request as well `final context = _MockRequestContext();`

erickzanardo: It mocks the request context only, I am mentioning the Request i.e 'context.request'.

My point is that, yes I agree with you, in your example where the route handler only return a body message, that when it does not read anything from the request, the proposed API in this PR provides little to none benefit.

But when you want to test a handler that needs the request to be mocked, with headers, method, body, etc. Then it could help the user to write tests with fewer code, since the testRouteHandler will mock everything for ya, you just pass the values.

felangel: Ah sorry I misunderstood. In that case, maybe `package:dart_frog_test` should focus on exporting classes like MockRequestContext, MockRequest, and utilities that make working with those classes easier rather than providing a custom test method?

erickzanardo: Yeah! That is an interesting idea, it will certainly make this package smaller

erickzanardo:
Closing in favor of #899

PR #908, opened 2023-08-08
feat(dart_frog): add optional shared parameter to serve function

femalemonkeyman:
add an optional shared parameter to the serve function to allow multi-threading of the resulting server

## Status

**Ready?**

## Description

add an optional shared parameter to the serve function to allow multi-threading of the resulting server

## Type of Change

- [x ] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


erickzanardo (dismissed):

packages/dart_frog/lib/src/serve.dart
```diff
@@ -26,5 +27,6 @@ Future<HttpServer> serve(
     port,
     poweredByHeader: poweredByHeader,
     securityContext: securityContext,
+    shared: shared,
```
alestiago, line 30: Can we add some tests to verify it defaults to false and that is being passed as expected? 

Inspiration can be taken from the test group [X-Powered-By-Header](https://github.com/VeryGoodOpenSource/dart_frog/blob/main/packages/dart_frog/test/src/serve_test.dart#L47) at the [serve_test.dart](https://github.com/VeryGoodOpenSource/dart_frog/blob/main/packages/dart_frog/test/src/serve_test.dart#L47) file.

alestiago (commented):
Thanks for contributing! LGTM, just need to add some tests 💙

packages/dart_frog/lib/src/serve.dart
femalemonkeyman: Would that be as simple as me checking the shared property on the server or would I check if multithreading it works properly?

alestiago: I think the former would be enough since the latter should be tackled by shelf.

femalemonkeyman: I have to say I feel a little silly but I can't seem to find a way to do the prior.
`      test('is configured by default', () async {
        try {
          final server = await serve((_) => Response(), 'localhost', 3000);
          final testServer = await serve((_) => Response(), 'localhost', 3000);
          await server.close();
          await testServer.close();
        } catch (e) {
          expect(
            e.toString(),
            equals(
              SocketException(
                'Failed to create server socket',
                osError: OSError(
                  'The shared flag to bind() needs to be `true` if binding multiple times on the same (address, port) combination.',
                ),
                address: (e as SocketException).address,
                port: 3000,
              ).toString(),
            ),
          );
        }
      });`
      Toying with this but even then I cannot close close the servers obbviously.

packages/dart_frog/test/src/serve_test.dart
```diff
@@ -90,6 +90,38 @@ void main() {
...
+        try {
+          servers
+            ..add(await serve((_) => Response(), 'localhost', 3000))
+            ..add(await serve((_) => Response(), 'localhost', 3000));
+        } catch (_) {}
+        for (final i in servers) {
+          await i.close();
+        }
+        expect(servers.length, 1);
+      });
```
alestiago, line 105:
```suggestion
    test(
          '''when false throws a SocketException when binding bind to the same combination of address and port''',
          () async {
        final server1 = await serve((_) => Response(), 'localhost', 3000);

        await expectLater(
          () async => serve((_) => Response(), 'localhost', 3000),
          throwsA(isA<SocketException>()),
        );

        await server1.close();
      });
```

packages/dart_frog/lib/src/serve.dart
```diff
@@ -16,6 +16,7 @@ Future<HttpServer> serve(
   int port, {
   String? poweredByHeader = 'Dart with package:dart_frog',
   SecurityContext? securityContext,
+  bool shared = false,
```
alestiago, line 19: Can we add some documentation to this parameter in the functions Dart doc?

I was thinking something like:

```dart
/// ...
/// If a [securityContext] is provided an HTTPS server will be started.
///
/// Enable [shared] to allow additional [HttpServer] objects to bind to the
/// same combination of [HttpServer.address] and [HttpServer.port].
```

packages/dart_frog/lib/src/serve.dart
alestiago: Thanks @femalemonkeyman for adding tests! Do not feel silly about this 💙 . I've added a few suggestions to the tests, let me know what you think 👀 🙌

packages/dart_frog/lib/src/serve.dart
femalemonkeyman: Thank you, this would be my first time doing something like this so I'm missing the proper etiquette

packages/dart_frog/test/src/serve_test.dart
```diff
@@ -90,6 +90,48 @@ void main() {
...
+    test(
+          '''when true serves requests successfully when binding bind to the same combination of address and port''',
+          () async {
+        final server1 = await serve(
+          (_) => Response(headers: {'server': '1'}),
+          'localhost',
+          3200,
+          shared: true,
+        );
+        final server2 = wait serve(
```
alestiago, line 116:
```suggestion
        final server2 = await serve(
```

alestiago (dismissed):

packages/dart_frog/test/src/serve_test.dart
```diff
@@ -90,6 +90,48 @@ void main() {
       });
     });
 
+    group('shared', () {
+    test(
```
alestiago, line 94: Can we run the formatter here? It doesn't seem to have proper indentation?

femalemonkeyman: Yeah that is weird

alestiago (approved):

wolfenrain (approved):

Issue #642, opened 2023-05-11
build: Share code between multiple dart_frog projects in a monorepo


Hannnes1:
**Description**

I'm looking for a way to share code between multiple `dart_frog` projects inside a monorepo, using local dependencies.
The current solution for sharing code between server and client seems to be to place the shared code inside the server project. (https://github.com/VeryGoodOpenSource/dart_frog/issues/410#issuecomment-1303707714) That solution works fine, since a Flutter project can import code outside of its own folder.

This solution does however not work when sharing code between two `dart_frog` projects, since one of the projects will always have to import from outside its folder (unless you nest all projects that depend on shared code).

I have thought about a few different possible solutions to this. The first is to use symlinks, but soft links don't work with Docker, and hard links don't work with git. The other option I can come up with is to be able to set a custom docker build context, since – as far as I understand it – that is the cause of this limitation.

I don't have an idea for how this could be implemented at a technical level, but it is possible to specify the docker file location and build context separately. For example:
`docker build ../ -f path/to/docker file`

That would allow for a file structure like this:
```
frog_1
frog_2
shared_code
client
```

Of course, this would require changing the |`Dockerfile` based on the build context and folder names. It would also be necessary to be able to specify the path to the build context in some way.

**Requirements**

- [ ] Allow importing from outside the project
- [ ] Allow specifying the docker build context

This issue is mostly intended as a starting point for a discussion of how this can be solved, and I am by no means saying that my proposed solution is the best one. I'm sure that the developers working on `dart_frog` have thought about this before. I would love to hear those thoughts, and if I have missed something.
This is quite an important feature for us, so I would be really happy if a solution can be found.

Hannnes1:
I haven't tested it myself yet, but has this been fixed by https://github.com/VeryGoodOpenSource/dart_frog/pull/910? I noticed that the issue was closed and then reopened.

alestiago:
Hi @Hannnes1 , I'm sorry for the confusion. This feature was fixed with #910, however the issue was reopened because it was yet to be released. You should now be able to use it with Dart Frog CLI latest version 🙌

Hannnes1:
Great, thanks!

PR #910, opened 2023-08-08
feat: allow prod server brick to use external dependencies

erickzanardo:
## Status

**IN DEVELOPMENT**

## Description

Allows the dart frog prod brick to build with external path dependencies.

Related to #642 

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


bricks/dart_frog_prod_server/hooks/src/create_external_packages_folder.dart
```diff
@@ -0,0 +1,99 @@
+import 'dart:io';
+
+import 'package:io/io.dart';
+import 'package:path/path.dart' as path;
+
+import 'get_path_dependencies.dart';
+
+Future<List<String>> createExternalPackagesFolder(Directory directory) async {
```
alestiago, line 8: Do we also consider getting external path dependencies of external path dependencies (recursive)?

alestiago:
This starts looking like a bundling process, do you think we can consider it to be its own brick/package/process instead of being baked in the Dart Frog Prod Server brick?

bricks/dart_frog_prod_server/hooks/src/create_external_packages_folder.dart
erickzanardo: oh yes, I didn't think on those cases, indeed we should! Good callout

erickzanardo:
> This starts looking like a bundling process, do you think we can consider it to be its own brick/package/process instead of being baked in the Dart Frog Prod Server brick?

The Dart Frog Prod Server brick is basically that already tbh, even before the change. I would refrain from adding a new brick for now, since this only needed here, maybe if this ever grow too much, but for now I would keep it as is.

bricks/dart_frog_prod_server/hooks/pre_gen.dart
```diff
@@ -21,11 +24,21 @@ Future<void> run(HookContext context) => preGen(context);
...
   io.Directory? directory,
+  ProcessRunner runProcess = io.Process.run,
   RouteConfigurationBuilder buildConfiguration = buildRouteConfiguration,
   void Function(int exitCode) exit = _defaultExit,
+  Future<void> Function(String from, String to) copyPath = io_expanded.copyPath,
 }) async {
   final projectDirectory = directory ?? io.Directory.current;
 
+  // We need to make sure that the pubspec.lock file is up to date
+  await dartPubGet(
```
alestiago, line 35: I think we should definitely allow the user to avoid performing the `dartPubGet`, maybe via a` --[no-]pub ` flag, similar to `flutter run --no--pub`. This flag would determine wether to run "dart pub get" before executing the generation, defaulting to on.

bricks/dart_frog_prod_server/hooks/pre_gen.dart
```diff
@@ -64,25 +77,23 @@ Future<void> preGen(
...
-    },
-  );
-
   final customDockerFile = io.File(
     path.join(projectDirectory.path, 'Dockerfile'),
   );
 
+  // Get all the internal path packages
+  final pathDependencies = (await getPathDependencies(projectDirectory))
+      .where((dependencyPath) => !dependencyPath.startsWith('..'))
```
alestiago, line 86: I would use the `path` package here. For a file path to be parent it doesn't always start with `..`, consider: `foo/../../`

bricks/dart_frog_prod_server/hooks/src/create_external_packages_folder.dart
alestiago: Awesome. I would personally avoid doing it recursive though, since a stack overflow could be reached. Instead, I would rather do a DFS style algorithm. Where the depth is increased until all packages are leafs or have already been resolved.

bricks/dart_frog_prod_server/hooks/pre_gen.dart
erickzanardo: I disagree, after my last commit, the brick now uses the pubspec lock file to get the external dependencies, since it provides us an accurate and reliable way to get all the external dependencies, even if they are external dependencies from external dependencies.

And like I added in the comment, we need to make sure that the pubspec.lock exists and is up to date.

Also, we already do a pub get in the generated bundle, with no option to skip it, because we do need the dependencies to be resolved to make the bundle, so I feel there is no real need for such flag.

alestiago: I feel the use case might apply in:
- Trying to use the command without internet connection
- Running on CI with a cached _pubspec.lock_

erickzanardo: We could, but I still find it dangerous. Let me exemplify why.

We could implement that flag, and do a validation, if you are passing the flag and there is no pubspec.lock, we give the user an error. That would cover the `Running on CI with a cached pubspec.lock` scenario that you mentioned.

But nothing guarantees that the pubspec.lock that is cached/checked under version control is up to date, and there is no way that we can check if it is (at least not without writing a quite complex code that would be potentially fragile) and give the user an reasonable error and then this can potentially break epically when bundling (this applies for the pub get that I introduced, as well to the pub get that is already being done at the end).

I still believe that this flag might introduce more danger, issues than benefits, so I would rather not consider adding it. But if feel otherwise, lets follow what you said and leave this for other PRs/versions :)

bricks/dart_frog_prod_server/hooks/pre_gen.dart
erickzanardo: good call, I replaced this elsewhere, was going to replace here, but seems that I forgot. I have changed to use path.

bricks/dart_frog_prod_server/hooks/src/create_external_packages_folder.dart
erickzanardo: Yeah, but the credits go for @wolfenrain, he mentioned this idea to me yesterday hahaha

bricks/dart_frog_prod_server/hooks/pre_gen.dart
```diff
@@ -64,25 +77,26 @@ Future<void> preGen(
...
+            path.isWithin(projectDirectory.path, dependencyPath),
+      )
+      .toList();
+
+  // Then create the external packages folder
+  // and add it to the list of path packages.
+  final externalDependencies = await createExternalPackagesFolder(
+    projectDirectory,
+    copyPath: copyPath,
+  );
```
alestiago, line 97: **Nit:** I feel the code already does a great job at explaining what it does.

```suggestion
  final externalDependencies = await createExternalPackagesFolder(
    projectDirectory,
    copyPath: copyPath,
  );
```

wolfenrain (approved):

bricks/dart_frog_prod_server/hooks/src/create_external_packages_folder.dart
```diff
@@ -0,0 +1,104 @@
...
+      ),
+    )..createSync();
+
+    // Copy all the dependencies to the packages directory
+    for (final entry in mappedDependencies.entries) {
+      final from = pathResolver.join(directory.path, entry.value);
+      final to = pathResolver.join(packagesDirectory.path, entry.key);
+
+      await copyPath(from, to);
+    }
```
alestiago, line 64: **Nit:** Simplify by using the already defined `pathDependencies`
```suggestion
    final dependencyOverrides = StringBuffer()
      ..writeln('dependency_overrides:');
    for (final dependencyPath in pathDependencies) {
      final basename = pathResolver.basename(dependencyPath);
      final from = pathResolver.join(directory.path, basename);
      final to = pathResolver.join(packagesDirectory.path, dependencyPath);

      await copyPath(from, to);

      dependencyOverrides.writeln('  $basename:\n    path: $to');
    }

    await File(
      pathResolver.join(
        buildDirectory.path,
        'pubspec_overrides.yaml',
      ),
    ).writeAsString(dependencyOverrides.toString());
```

bricks/dart_frog_prod_server/hooks/src/create_external_packages_folder.dart
```diff
@@ -0,0 +1,104 @@
...
+    );
+
+    await File(
+      pathResolver.join(
+        buildDirectory.path,
+        'pubspec_overrides.yaml',
+      ),
+    ).writeAsString('''
+dependency_overrides:
+${mappedPaths.entries.map((entry) => '  ${entry.key}:\n    path: ${entry.value}').join('\n')}
```
alestiago, line 83: Here we are using `{entry.key}` which resolves to the basename of the dependency path. I don't think this might be always true, instead we should use the name provided by the entry in the `pubspec.lock` or `pubspec.yaml`. Since we are already reading the `pubspec.lock` I think the former makes more sense.

alestiago (commented):
LGTM.

Only one "worrying" observation, which is the usage of the basename of the dependency path as the package name instead of the one given by the `pubspec`.

bricks/dart_frog_prod_server/hooks/src/create_external_packages_folder.dart
erickzanardo: I would rather keep the operations separated, we definitely could put them together, but I feel the for would end up a little bit mixed in its responsabilities, and I would rather have the code easier to read.

If you feel too strong in the other direction, let me know and I can revisit.

bricks/dart_frog_prod_server/hooks/pre_gen.dart
```diff
@@ -64,25 +77,30 @@ Future<void> preGen(
...
```
alestiago, line 97: **Nit**: I personally think the code is already self-explanatory and it is doing a fantastic job at describing what is being done. Unlike code, comments do not get updated as code changes and thus they might not always tell the truth about the code. Hence, if the code is already clear I would personally disregard the comments. 

These thoughts are not mine, they have been stolen from Uncle's Bob book, I think this [document](https://bpoplauschi.github.io/2021/01/20/Clean-Code-Comments-by-Uncle-Bob-part-2.html) cites some of it.

```suggestion
  final internalPathDependencies = (await getPathDependencies(projectDirectory))
      .where(
        (dependencyPath) =>
            path.isWithin(projectDirectory.path, dependencyPath),
      )
      .toList();

  final externalDependencies = await createExternalPackagesFolder(
    projectDirectory,
    copyPath: copyPath,
  );
```

bricks/dart_frog_prod_server/hooks/pre_gen.dart
```diff
@@ -64,25 +77,30 @@ Future<void> preGen(
...
     path.join(projectDirectory.path, 'Dockerfile'),
   );
 
+  // Get all the internal path packages
+  final internalPathDependencies = (await getPathDependencies(projectDirectory))
+      .where(
+        (dependencyPath) =>
+            path.isWithin(projectDirectory.path, dependencyPath),
+      )
+      .toList();
```
alestiago, line 90: **Nit**: Does it really need to be a list or can it be an iterable?

```suggestion
      );
```

bricks/dart_frog_prod_server/hooks/src/create_external_packages_folder.dart
```diff
@@ -0,0 +1,100 @@
+import 'dart:io';
+
+import 'package:io/io.dart' as io;
+import 'package:path/path.dart' as path;
+import 'package:pubspec_lock/pubspec_lock.dart';
+
+Future<List<String>> createExternalPackagesFolder(
```
alestiago, line 7: **Nit:** Does the order matter?

```suggestion
Future<Iterable<String>> createExternalPackagesFolder(
```

alestiago (approved):
LGTM! Thanks for taking care of this. Looking forward to seeing the follow-up PR 💙

bricks/dart_frog_prod_server/hooks/pre_gen.dart
```diff
@@ -64,25 +77,26 @@ Future<void> preGen(
...
-      exit(1);
-    },
-  );
-
   final customDockerFile = io.File(
     path.join(projectDirectory.path, 'Dockerfile'),
   );
 
+  // Get all the internal path packages
+  final pathDependencies = (await getPathDependencies(projectDirectory))
```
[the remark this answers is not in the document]
erickzanardo: Will work on that in the following PR as discussed :)

wolfenrain (approved):

erickzanardo:
@alestiago will address some of the resolved comments in the follow up!

PR #915, opened 2023-08-10
feat(vscode): implement daemon protocol

alestiago:
## Status

**READY**

## Description

Changes:
- Implements base classes that adhere to the Dart Frog deamon protocol.
- Implements functionality to decode messages

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


extensions/vscode/src/daemon/protocol.ts
```diff
@@ -0,0 +1,37 @@
+/**
+ * Defines the protocol used by the Dart Frog daemon and custom
+ * type guards to check if an object is a valid message.
+ *
+ * @see {@link https://github.com/VeryGoodOpenSource/dart_frog/blob/main/packages/dart_frog_cli/lib/src/daemon/protocol.dart Dart Frog dart's protocol}
+ */
+
+export interface DaemonMessage {}
```
erickzanardo, line 8: So we really need this interface? can't we just use the abstract class instead?

alestiago: It is useful for multiple reasons:
- Formally matches directly with the [protocol](https://github.com/VeryGoodOpenSource/dart_frog/blob/main/packages/dart_frog_cli/lib/src/daemon/protocol.dart) and its hierarchy. 
- Useful as fallback type when decoding deamon messages from stdout (better than falling back to Object or any)
- Arguably easier when declaring instances of iterables with messages

alestiago: I've now included the method to decode messages in this PR to avoid confusion.

extensions/vscode/src/daemon/protocol.ts
```diff
@@ -0,0 +1,59 @@
...
+  /**
+   * Decodes messages that follow the protocol.
+   *
+   * @param data The data to decode (usually from stdout of the Dart Frog
+   * daemon and in JSON format).
+   * @returns The decoded messages.
+   */
+  public static decode(data: Buffer): DaemonMessage[] {
+    const stringData = data.toString();
+    const messages = stringData.split("\n").filter((s) => s.trim().length > 0);
```
wolfenrain, line 18: Can't we use the `toJSON` method?

extensions/vscode/src/daemon/protocol.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+}
+
+export abstract class DaemonRequest implements DaemonMessage {
+  abstract method: string;
+  abstract id: string;
+  abstract params: any;
+}
+
+export function isDeamonRequest(object: any): object is DaemonRequest {
```
wolfenrain, line 38: Deamon or Daemon 😏

erickzanardo (approved):

wolfenrain (approved):

PR #926, opened 2023-08-16
feat: change production brick to support external dependencies in the Dockerfile

erickzanardo:
## Status

**READY**

## Description

This PR change the production brick to be able to support external path dependencies.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


bricks/dart_frog_prod_server/__brick__/build/{{#addDockerfile}}Dockerfile{{/addDockerfile}}
```diff
@@ -3,13 +3,20 @@
...
+{{#hasExternalDependencies}}
 
+# Copy external dependencies
+COPY ./.dart_frog_path_dependencies  ./.dart_frog_path_dependencies
+COPY ./pubspec_overrides.yaml ./pubspec_overrides.yaml
+{{/hasExternalDependencies}}
 # Copy Dependencies
 {{#pathDependencies}}COPY {{{.}}} ./{{{.}}}
 {{/pathDependencies}}
+{{^hasExternalDependencies}}
```
erickzanardo, line 15: Sadly, we need to give up on these in between docker cache when having external dependencies.

Since internal path dependencies can depend on external path dependencies and the code that bundles the external dependencies, bundles then at the root, when individually "pub getting" an internal path dependency, the process will fail.

The alternative would be to recursively run a bundling for all the internal dependencies, but IMO this will lead our brick to be too complex and harder to maintain, so I decided that when we have external path dependencies, the only pub get that will be executed, will be on the root pubspec, we will loose the in between layers caching from docker, but the tradeoff in this case is justified.

felangel (approved):

PR #953, opened 2023-08-23
feat(vscode): allow stopping development server

alestiago:
## Status

**READY**

## Description

Allows stopping a development server via the command palette.

Changes:

- Registers stop-dev-server command

## Demonstration

https://github.com/VeryGoodOpenSource/dart_frog/assets/44524995/c63ca509-370d-4b0e-924f-c7e997df439d

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


extensions/vscode/src/commands/stop-dev-server.ts
```diff
@@ -0,0 +1,190 @@
...
+ * Stops a Dart Frog development server on the current workspace.
+ *
+ * This command can be launched from the command palette.
+ *
+ * The user requires to have Dart Frog CLI installed in order to run this
+ * command. If they opt out of installing Dart Frog CLI, the command will
+ * run at their own risk.
+ *
+ * If there are no running servers or the daemon is not ready, the user will
+ * be prompted to start a server and this command will not run.
```
erickzanardo, line 27: I wonder if this command shouldn't just be a no op in this case. If you want to stop a server, but none is running, what is the reasoning for starting it?

extensions/vscode/src/commands/stop-dev-server.ts
```diff
@@ -0,0 +1,190 @@
...
+
+      await deregisterPromise;
+
+      progress.report({
+        message: `Server stopped successfully`,
+        increment: 100,
+      });
+
+      // Add a small delay to allow the user to read the message.
+      return await new Promise((resolve) => setTimeout(resolve, 250));
```
erickzanardo, line 103: Nit: 250 millis to little to me hahaha, maybe at least a second?

extensions/vscode/src/commands/stop-dev-server.ts
```diff
@@ -0,0 +1,190 @@
...
+    (application) => new PickableDartFrogApplication(application)
+  );
+  quickPick.show();
+
+  return new Promise<DartFrogApplication | undefined>((resolve) => {
+    quickPick.onDidChangeSelection((value) => {
+      quickPick.dispose();
+
+      if (value.length === 0) {
+        resolve(undefined);
```
erickzanardo, line 155:
```suggestion
        resolve();
```

`undefined` is already implicity

extensions/vscode/src/commands/stop-dev-server.ts
alestiago: It is somewhat of a no-op when there are no servers running (it does nothing). I see it as a bad UX if no feedback is given when the command is executed and that's why I do prefer showing at least some information that there are no servers running.

I have updated the documentation since it might have been ambiguos before.

extensions/vscode/src/commands/stop-dev-server.ts
alestiago: Did you try it out? I felt 1000 seconds looked too much and the progress animation doesn't feel right with such amount.

extensions/vscode/src/commands/stop-dev-server.ts
alestiago: This can't be done since the resolve function expects one argument.

erickzanardo: hmmm, got ya, TypeScript must define the types that way then, which makes no sense to me, since you can have a void Promise, this argument should be optional 🤔

erickzanardo (approved):

PR #1025, opened 2023-09-11
feat(vscode): implemented quickPickProject

alestiago:
## Status

**READY**

## Description

Adds an option to quick pick projects. This is done in preparation of #1011. The idea is that when there is more than a single Dart Frog project found on the monorepo this quick pick will be shown.

Changes:
- Adds `quickPickProject`

## Type of Change

- [X] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

PR #1026, opened 2023-09-12
feat(vscode): show quickPickProject on monorepos

### LINKED ISSUE #1011: feat(vscode): support running on monorepos
## Problem

Currently, the extension detects a Dart Frog project by traversing the file system upwards until it finds a Dart Frog project. If no Dart Frog project is found the vast majority of UI elements provided by the Dart Frog extension are hidden to avoid unnecessarily polluting the user's workspace.

By design, we originally expected users to open a workspace at the root of the Dart Frog project. However, some users do not do so because their workspace is a monorepo with one or more Dart Frog project inside it. This not only hides the Dart Frog's extension UI for them but also running commands such as "Start Development Server" will report: 

>  Failed to find a Dart Frog project within the current workspace.

Currently, the extension can run in a monorepo but this **requires the user to have an active route file open**.

## Proposal

There are multiple solutions, but after discussing with a closed group of users the preferred solution was to:
1. Traverse the file system downwards to find Dart Frog projects after failing to traverse upwards.
2. If more than one Dart Frog project is found when traversing downwards, then upon starting a server allow the user to pick their preferred project every single time. If the user finds the prompt annoying, then they should open the preferred Dart Frog project from the root.

## Requirements

- [ ] Allow running the extension on monorepos
- [ ] Show user's a quick-pick prompt when there is more than one Dart Frog project on the monorepo and they attempt to run a command that requires a Dart Frog project to be specified

alestiago:
> Should be merged after #1025 

## Status

**READY**

## Description

Resolves #1011 

Changes:
- Updates `resolveDartFrogProjectPathFromWorkspaceFolders` to use `nearestChildDartFrogProjects`
- Shows `quickPickProject` when there is more than one Dart Frog project resolved

## Demonstration

https://github.com/VeryGoodOpenSource/dart_frog/assets/44524995/87b016c5-ca4d-4edb-af67-7d69efbaa522

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


extensions/vscode/src/commands/new-route.ts
```diff
@@ -49,8 +50,21 @@ export const newRoute = async (uri: Uri | undefined): Promise<void> => {
   if (uri === undefined) {
     selectedPath = resolveDartFrogProjectPathFromActiveTextEditor();
 
-    if (selectedPath === undefined) {
-      selectedPath = resolveDartFrogProjectPathFromWorkspaceFolders();
+    if (!selectedPath) {
```
erickzanardo, line 53: This snippet, and the above are identical if I am not mistaken, can we wrap it in a method or function so we don't repeat ourselves?

extensions/vscode/src/commands/start-dev-server.ts
```diff
@@ -57,13 +58,28 @@ export const startDevServer = async (): Promise<
...
   }
 
-  let workingPath = resolveDartFrogProjectPathFromWorkspaceFolders();
-  if (workingPath === undefined) {
-    workingPath = resolveDartFrogProjectPathFromActiveTextEditor();
+  let projectPath: string | undefined;
+  const dartFrogProjectsPaths =
+    resolveDartFrogProjectPathFromWorkspaceFolders();
+
+  if (dartFrogProjectsPaths) {
```
erickzanardo, line 65: Here as well

extensions/vscode/src/commands/new-route.ts
alestiago: This is a valid suggestion. I do plan to refactor to reuse logic, if you don't mind I'll do the refactoring in another Pull Request 🙌

alestiago: I've simplified the logic, let me know what you think now 🙌

alestiago: I've simplified the logic, let me know what you think now 🙌

renancaraujo (approved):

erickzanardo (approved):

Issue #745, opened 2023-07-10
fix: Improve dependency injection order


rubenferreira97:
**Description**

The current order of dependency injection, where dependencies are injected from bottom to top, feels unnatural and counterintuitive.

**Steps To Reproduce**
```dart
// routes/_middleware.dart
Handler middleware(Handler handler) {

  /* Gives exception
  return handler
      .use(provider<int>((_) => 1))
      .use(provider<String>((context) => context.read<int>().toString()));
  */

  // Runs fine
  return handler
      .use(provider<String>((context) => context.read<int>().toString()))
      .use(provider<int>((_) => 1));
}

// routes/index.dart
Response onRequest(RequestContext context) {
  return Response(body: context.read<String>()); // Should print '1'
}
```

**Expected Behavior**
Dependency injection should follow a sequential order for proper execution. The following code should work without issues:
```dart
Handler middleware(Handler handler) {
  return handler
      .use(provider<int>((_) => 1))
      .use(provider<String>((context) => context.read<int>().toString()));
}

Response onRequest(RequestContext context) {
  return Response(body: context.read<String>()); // Should print '1'
}
```

And the following IMO should fail:

```dart
Handler middleware(Handler handler) {
  return handler
      .use(provider<String>((context) => context.read<int>().toString()))
      .use(provider<int>((_) => 1));
}

Response onRequest(RequestContext context) {
  return Response(body: context.read<String>()); // Should print '1'
}
```

alestiago:
Hi @quentin7b , I believe this change would probably be something significant on how users structure their code, so the team might need some time to evaluate it further. However, adding more documentation around this behaviour sounds like a good idea to me. If you're interested in contributing let us know and we'll be happy to review your Pull Request.

quentin7b:
Hi !
Of course I can add a specific section in the https://dartfrog.vgv.dev/docs/basics/dependency-injection page that describes this behavior and mention this issue ? 
What do you think about this ?

alestiago:
Hi @quentin7b , that sounds good to me. It will be great if you can describe the current behaviour and include examples that show and build the readers intuition around it. Give it a try and we'll happily review it! 🙌

rubenferreira97:
Even if documentation is added (which is an improvement), I still think it would be nice to address this. Yes, it is a breaking change, but I believe this is a kind of issue for newcomers.

alestiago:
[...]

PR #1218, opened 2023-12-19
docs(dependency injection): Add a section about ordering providers

quentin7b:
## Status

**READY**

## Description

As discussed in #745, I've added a specific section in the *dependency injection* page of the documentation specifying that `dart_frog` uses `bottom to top` order in middlewares.

To make it as clear as possible, I've added a basic example of a repository that depends on a database connection. So the database connection provider should be set after the repository.

I've also added a note that links to the #745 issue

## Type of Change

- [ ] 🧹 Code refactor
- [X] 📝 Documentation
- [ ] 🗑️ Chore


wolfenrain:
Hey @quentin7b thanks for the PR and sorry for the delayed response (holiday season and such), I'll be reviewing your PR today!

docs/docs/basics/dependency-injection.md
```diff
@@ -158,3 +158,106 @@ Middleware cachedAsyncGreetingProvider() {
 ```
 
 With the above implementations, the greeting will only be computed once and the cached value will be used for the duration of the application lifecycle.
+
+### Order matter
+
+In a real life application you will find yourself adding multiple `providers` to your project.
+
+Also, some `providers` **will depend** on others, as in any application relying on dependency injection.
```
wolfenrain, line 166:
```suggestion
Some `providers` **will depend** on others, as with any application relying on dependency injection.
```

docs/docs/basics/dependency-injection.md
```diff
@@ -158,3 +158,106 @@ Middleware cachedAsyncGreetingProvider() {
...
+
+Here is an example with two `providers`
```
wolfenrain, line 168:
```suggestion
Here is an example with two `providers` where the second one depends on the first one for it to work:
```

docs/docs/basics/dependency-injection.md
```diff
@@ -158,3 +158,106 @@ Middleware cachedAsyncGreetingProvider() {
...
+
+At this point, it seems clear that `myItemsDataSourcesMiddlewareProvider` depends on `databaseConnectionMiddlewareProvider`.
+
+When you'll try to access the instance of `ItemDataSource` using `context.read<ItemDataSource>()`, here is what will happen:
+
+1. _dart_frog_ will try to create the instance and return it. To do so, it will create a `DatabaseItemDataSource` object and to fulfill its `connection` parameter
+2. It will look "above" in the dependency graph for a provider of `Connection`
+3. It will find it with `databaseConnectionMiddlewareProvider`, and so on.
+
+This is how DI works, but the question is "what look "above" means?"
```
wolfenrain, line 218:
```suggestion
This is how dependency injections works, but let's clarify what "look "above" means.
```

docs/docs/basics/dependency-injection.md
```diff
@@ -158,3 +158,106 @@ Middleware cachedAsyncGreetingProvider() {
...
+
+It means that you have to tell _dart_frog_ how to build a `Connection` **before** how to build a `ItemDataSource`.
+
+How do we do so? By ordering the _Providers_.
```
wolfenrain, line 222:
```suggestion
We can tell Dart Frog how to build a `Connection` **before** it builds a `ItemDataSource`. We do that by defining the order of the providers.
```

docs/docs/basics/dependency-injection.md
```diff
@@ -158,3 +158,106 @@ Middleware cachedAsyncGreetingProvider() {
...
+
+:::tip
+In _dart_frog_, dependencies are resolved from **bottom** to **top**
```
wolfenrain, line 225:
```suggestion
In Dart Frog, dependencies are resolved from **bottom** to **top**  
```

docs/docs/basics/dependency-injection.md
```diff
@@ -158,3 +158,106 @@ Middleware cachedAsyncGreetingProvider() {
...
+
+It means that you have to tell _dart_frog_ how to build a `Connection` **before** how to build a `ItemDataSource`.
+
+How do we do so? By ordering the _Providers_.
+
+:::tip
+In _dart_frog_, dependencies are resolved from **bottom** to **top**  
+:::
+
+So if `B` depends on `A`, make sure you declare them this way
```
wolfenrain, line 228:
```suggestion
So if provider `B` depends on provider `A`, you will have to declare them as followed:
```

docs/docs/basics/dependency-injection.md
```diff
@@ -158,3 +158,106 @@ Middleware cachedAsyncGreetingProvider() {
...
+
+```dart
+Handler middleware(Handler handler) {
+  return handler
+      .use(B())
+      .use(A())
+}
+```
+
+In our example, `myItemsDataSourcesMiddlewareProvider` depends on `databaseConnectionMiddlewareProvider`, so this will work
```
wolfenrain, line 238:
```suggestion
In the example we gave at the beginning where `myItemsDataSourcesMiddlewareProvider` depends on `databaseConnectionMiddlewareProvider` we know it will work when we provided them like so:
```

docs/docs/basics/dependency-injection.md
```diff
@@ -158,3 +158,106 @@ Middleware cachedAsyncGreetingProvider() {
...
+
+```dart
+Handler middleware(Handler handler) {
+  return handler
+      .use(myItemsDataSourcesMiddlewareProvider())
+      .use(databaseConnectionMiddlewareProvider())
+}
+```
+
+but this won't !
```
wolfenrain, line 248:
```suggestion
But if we change the order of the providers it will not work:
```

wolfenrain (commented):
I never submitted my comments so here they are 😂 

cc: @Mayaa-s can you have a look as well?

quentin7b:
Thank you for the comments @wolfenrain, I'll wait for @Mayaa-s to have a look to do the edits !

docs/docs/basics/dependency-injection.md
```diff
@@ -158,3 +158,106 @@ Middleware cachedAsyncGreetingProvider() {
...
+}
+```
+
+:::note
+You can find a complete example of data source implementation in the [todos tutorial](/docs/tutorials/todos)
+:::
+
+At this point, it seems clear that `myItemsDataSourcesMiddlewareProvider` depends on `databaseConnectionMiddlewareProvider`.
+
+When you'll try to access the instance of `ItemDataSource` using `context.read<ItemDataSource>()`, here is what will happen:
```
Mayaa-s, line 212:
```suggestion
When you try to access the instance of `ItemDataSource` using `context.read<ItemDataSource>()`, here is what will happen:
```

Mayaa-s (commented):
Looks good with Jochum's edits!

docs/docs/basics/dependency-injection.md
```diff
@@ -158,3 +158,104 @@ Middleware cachedAsyncGreetingProvider() {
 ```
 
 With the above implementations, the greeting will only be computed once and the cached value will be used for the duration of the application lifecycle.
+
+### Order matter
```
alestiago, line 162:
```suggestion
### Order matters
```

docs/docs/basics/dependency-injection.md
```diff
@@ -158,3 +158,104 @@ Middleware cachedAsyncGreetingProvider() {
...
+
+At this point, it seems clear that `myItemsDataSourcesMiddlewareProvider` depends on `databaseConnectionMiddlewareProvider`.
+
+When you try to access the instance of `ItemDataSource` using `context.read<ItemDataSource>()`, here is what will happen:
+
+1. _dart_frog_ will try to create the instance and return it. To do so, it will create a `DatabaseItemDataSource` object and to fulfill its `connection` parameter
+2. It will look "above" in the dependency graph for a provider of `Connection`
+3. It will find it with `databaseConnectionMiddlewareProvider`, and so on.
+
+This is how dependency injections works, but let's clarify what "look "above" means.
```
alestiago, line 218:
```suggestion
2. It will "look above" in the dependency graph for a provider of `Connection`
3. It will find it with `databaseConnectionMiddlewareProvider`, and so on.

This is how dependency injections works, but let's clarify what "look above" means.
```

docs/docs/basics/dependency-injection.md
```diff
@@ -158,3 +158,104 @@ Middleware cachedAsyncGreetingProvider() {
...
+      .use(databaseConnectionMiddlewareProvider())
+}
+```
+
+But if we change the order of the providers it will not work:
+
+```dart
+Handler middleware(Handler handler) {
+  return handler
+      // This won't work because dart frog is bottom top
```
alestiago, line 251:
```suggestion
      // This won't work because Dart Frog is bottom top
```

docs/docs/basics/dependency-injection.md
```diff
@@ -158,3 +158,104 @@ Middleware cachedAsyncGreetingProvider() {
...
+      .use(databaseConnectionMiddlewareProvider())
+      .use(myItemsDataSourcesMiddlewareProvider())
+}
+```
+
+:::note
+Right now, there is an issue about this [fix: Improve dependency injection order #745](https://github.com/VeryGoodOpenSource/dart_frog/issues/745) because some other DI frameworks are working top to bottom and dart_frog is bottom to top.
+
+So you might want to keep that in mind for future releases, but as it would be a breaking change, the release version will reflect this.
+:::
```
alestiago, line 261: 💯

alestiago (dismissed):
Thanks for adding this!

Overall looks good to me, just highlighted some small things.

The examples are good. I wonder if we could use dummy classes that would be easier to digest. Like for example, a `Car` depending on a `Wheel` (since it doesn't require knowledge about what a database connection or data source is). However, I'm happy as is too!

quentin7b:
> Thanks for adding this!
> 
> Overall looks good to me, just highlighted some small things.
> 
> The examples are good. I wonder if we could use dummy classes that would be easier to digest. Like for example, a `Car` depending on a `Wheel` (since it doesn't require knowledge about what a database connection or data source is). However, I'm happy as is to!

I think it is a good idea. It's true I've taken this example... Because that is my use case !
But Car and Wheel are better for understanding.
I'll update the PR soon with this changes

quentin7b:
> > Thanks for adding this!
> > Overall looks good to me, just highlighted some small things.
> > The examples are good. I wonder if we could use dummy classes that would be easier to digest. Like for example, a `Car` depending on a `Wheel` (since it doesn't require knowledge about what a database connection or data source is). However, I'm happy as is to!
> 
> I think it is a good idea. It's true I've taken this example... Because that is my use case ! But Car and Wheel are better for understanding. I'll update the PR soon with this changes

@alestiago the update with car factory has been done !

alestiago (approved):
Thank you for this contribution! 💙

wolfenrain (approved):

PR #1262, opened 2024-01-28
fix(dart_frog): allow access of `shelf.Message.context`

### LINKED ISSUE #773: feat: Allow routing parameters in `middleware`
**Description**
Currently is not possible to access routing parameters on middlewares. Would be nice in cases where we have nested routing parameters and have different logic dependent on that parameter (e.g. `/[userId]/tasks/[taskId]`, middleware on `/[userId]/_middleware.dart` that checks if we have permissions for that `[userId`) .

My current workaround relies on private implementations 😕:

**Middleware `/[test]/_middleware.dart`:**
```dart
import 'package:dart_frog/src/_internal.dart' show toShelfHandler;

Handler middleware(Handler handler /* String test */) {
  return fromShelfHandler((request) async {
    print((request.context['shelf_router/params'] as Map<String, String>?)?['test']);
    return toShelfHandler(handler)(request);
  });
}
```

**Index `/[test]/index.dart`:**
```dart
Future<Response> onRequest(RequestContext context, String test) async {
  return Response(body: '$test');
}
```

sun-jiao:
## Status

READY

## Description

Allow access of `shelf.Message.context`, which is used by some shelf middleware and shelf handlers. 
closes #773.

## Type of Change

- [x] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/dart_frog/lib/src/request.dart
+++ b/packages/dart_frog/lib/src/request.dart
@@ -100,6 +100,9 @@ class Request {
     return _request.context['shelf.io.connection_info']! as HttpConnectionInfo;
   }
 
+  /// Shelf context that can be used by shelf middleware and shelf handlers.
+  Map<String, Object> get shelfContext => _request.context;
+
   /// The requested url relative to the current handler path.
   Uri get url => _request.url;
 
--- a/packages/dart_frog/lib/src/response.dart
+++ b/packages/dart_frog/lib/src/response.dart
@@ -87,6 +87,9 @@ class Response {
   /// The HTTP status code of the response.
   int get statusCode => _response.statusCode;
 
+  /// Shelf context that can be used by shelf middleware and shelf handlers.
+  Map<String, Object> get shelfContext => _response.context;
+
   /// The HTTP headers with case-insensitive keys.
   /// The returned map is unmodifiable.
   Map<String, String> get headers => _response.headers;


tomarra:
Hi @sun-jiao 👋 thanks for opening this PR. It looks like there is an error in the CI pipeline due to missing tests. Can you please take a look and get this passing? Then I can work with the team to get feedback/reviews. Thanks in advance!

sun-jiao:
@tomarra Testcases are already added.

sun-jiao:
@tomarra any updates?

wolfenrain:
Hi thanks for the PR! I had a quick look and I think we should not expose the shelf context like this. Dart Frog tries to stay agnostic to it's underlying framework where it can and exposing this context would counter that.

Based on the example you gave I presume you wanted to have access to the route params in the middleware. I think we should rather focus on that use-case than exposing all of the context as a map. We could potentially expose those parameters in the the same way that we do with the `onRequest` method. 

```dart
  return (context, String routeParam1, String routeParam2) {
  
    // ... code ...
    
    return handler(context);
  }
}
```

This would require us to tweak how we build the routing pipeline but it should be doable and would bring the API more inline with the rest of Dart Frog. Any thoughts on that?

cc: @felangel

sun-jiao:
@wolfenrain No, currently I am not concerned with the router's middleware parameters. I'm developing a library (in fact, based on another shelf library) which provides session persistence for shelf. And I want to add support for frog. While it saves the session in shelf's `request.context`. This is the library: https://github.com/sun-jiao/session_shelf/tree/dart_frog.

sun-jiao:
@wolfenrain Hi, any updates?

sun-jiao:
Hello, any updates? Or is there any more elegant way to implement the same function?

wolfenrain:
> Hello, any updates? Or is there any more elegant way to implement the same function without exposing shelf context? @wolfenrain
> 
> Sessions are put in `shelf.Context` in my code: [sun-jiao/session_shelf@`dart_frog`/lib/middleware/cookies_middleware.dart#L67C22-L67C34](https://github.com/sun-jiao/session_shelf/blob/dart_frog/lib/middleware/cookies_middleware.dart?rgh-link-date=2024-03-10T08%3A31%3A06Z#L67C22-L67C34)
> 
> In fact, I just did a search before opening this PR to confirm if there was a duplicate, and found issue #773 occasionally.

Hi! Sorry for the late reply I'll be looking into what you linked later this week, preferably we would not want to expose that kind of shell stuff so I'll have a look at your use-case to see if we can come up with a different approach, otherwise we can always expose it as a `@visibleForTesting` so you could use it without us promoting for normal use too much

sun-jiao:
@wolfenrain Thanks.

alestiago:
@sun-jiao we're also have a Dart Plugin System on the works that would probably help you improve the developer experience for the users of your tool that rely on Dart Frog. Keep an eye on the next updates 👀

Issue #1327, opened 2024-03-28
feat: `dart_frog build -p` (specifies port) 


a-wallen:
**Description**

GCP allows sidecar images. The ingress image must handle requests on 8080, so that port is reserved. 

My sidecar image needs a liveness probe. 

**It can't also run on 8080**.

So, I want to be able to call `dart_frog build -p 8181`.

Looks like the main function reads from `Platform.environment` already.
```dart
void main() async {
  final address = InternetAddress.anyIPv6;
  final port = int.tryParse(Platform.environment['PORT'] ?? '8080') ?? 8080;
  createServer(address, port);
}
```

The easiest solution is for the `build` command handler to set the PORT variable.

IIUC another solution would be to do it in the Dockerfile.

Another (not preferred IMO) is to modify the generated `server.dart` code.

**Requirements**

- [ ] `dart_frog build -p <port>` runs the server on a port.

PR #1346, opened 2024-04-14
Added port number feature to build command

plato79:
## Status

READY

## Description

Added the port number to the build option for #1327

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


tomarra:
Hi @plato79 👋thanks for taking on the issue and opening a PR! It looks like there are some failures in CI so please take a look at the output and get everything to green before we start the review process. Thanks!

plato79:
Now, it's past anlalyze.. Though there is an error in prod_server_builder test. This one is a bit complex though, while the fix is very easy.

The code below is the reason for the error.

`prod_server_builder.dart:`
```(dart)
    final vars = <String, dynamic>{
      'dartVersion': dartVersion,
      'port': port,
    };

    logger.detail('[codegen] running pre-gen...');
    await prodServerBundleGenerator.hooks.preGen(
      vars: vars,
      workingDirectory: workingDirectory.path,
      onVarsChanged: (v) => vars..addAll(v ?? {}),
    );
```

There is a vars variable which has default `dartVersion` which is `"stable"` and default `port` which is `"8080"`. There is possibility additional variables are provided through mason. We add those to the current variables defined in `vars` we have here. It modifies the value if the same key is used so everything is alright.

...until no variables are sent to here. Which makes `v` `null` (it says it shouldn't happen but it happens). Why `v`has `null` value I am not sure. I thought that code would execute only when there is a value in `v` but this is not the situation.

Now the fix is easy as you can see. Just adding a null check fixes. But this time flutter says I shouldn't add a null check to a variable which shouldn't be null at all. While the code above fixes the problem and causes no problem, it probably will pop up in the ci analyze section as error just like the previous `"vars" should be defined as final" warning appears as error in analyze.

Do you have any suggestion about this?

plato79:
> Now, it's past anlalyze.. Though there is an error in prod_server_builder test. This one is a bit complex though, while the fix is very easy.
> 
> ~~The code below is the reason for the error.~~
> 
> `prod_server_builder.dart:`
> 
> ```
>     final vars = <String, dynamic>{
>       'dartVersion': dartVersion,
>       'port': port,
>     };
> 
>     logger.detail('[codegen] running pre-gen...');
>     await prodServerBundleGenerator.hooks.preGen(
>       vars: vars,
>       workingDirectory: workingDirectory.path,
>       onVarsChanged: (v) => vars..addAll(v ?? {}),
>     );
> ```
> 
>~~ There is a vars variable which has default `dartVersion` which is `"stable"` and default `port` which is `"8080"`. There is possibility additional variables are provided through mason. We add those to the current variables defined in `vars` we have here. It modifies the value if the same key is used so everything is alright.~~
> 
>~~ ...until no variables are sent to here. Which makes `v` `null` (it says it shouldn't happen but it happens). Why `v`has `null` value I am not sure. I thought that code would execute only when there is a value in `v` but this is not the situation.~~
> 
> ~~Now the fix is easy as you can see. Just adding a null check fixes. But this time flutter says I shouldn't add a null check to a variable which shouldn't be null at all. While the code above fixes the problem and causes no problem, it probably will pop up in the ci analyze section as error just like the previous `"vars" should be defined as final" warning appears as error in analyze.~~
> 
> Do you have any suggestion about this?~~

Ok, that was not the cause. Still checking to see what the real problem here is.

plato79:
This last commit should fix the error in the test.

plato79:
The failing tests are actually runs successfully if you execute the commands outside of the test. For example build_test fails, but if you run `dart run build/bin/server.dart` in the temporary directory it created and call `http://localhost:8080` it works. But the test itself cannot get a response from that page.

@tomarra Are you sure that these tests worked without problem before I made the change?

plato79:
I think this is ready for merge. What do you think?

alestiago:
Hi @plato79! I've added this to the review column for May 18 to May 31. Hopefully we get to review it by then 🙌

alestiago:
Hi @plato79 this seems silent for a while, are you still able to complete the effort?

plato79:
Sorry, because of my "paying" work, I couldn't concentrate on this for a while. I'll try to fix it tomorrow.

wolfenrain:
> Sorry, because of my "paying" work, I couldn't concentrate on this for a while. I'll try to fix it tomorrow.

Hey @plato79 have you had some time to look into this PR again?

PR #1410, opened 2024-06-22
perf: optimize `Request.read`

kevmoo:
## Status

**READY**

## Description

Use `Expando` instead of cloning each request.
- *Should* improve performance. Should be measured.
- Allows nested `_request` to be final. (Potentially allows future use of `extension type`

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


kevmoo:
@felangel – enable CI workflows?

felangel:
> @felangel – enable CI workflows?

I’m not part of the organization anymore so I don’t have permissions unfortunately. @wolfenrain @alestiago can help 👍

packages/dart_frog/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 1.1.1-wip
+
+- chore: optimize `Request.read()`.
+
```
wolfenrain, line 4:
```suggestion
```

We generate our changelogs 😉

packages/dart_frog/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: dart_frog
 description: A fast, minimalistic backend framework for Dart built by Very Good Ventures.
-version: 1.1.0
+version: 1.1.1-wip
```
wolfenrain, line 3:
```suggestion
version: 1.1.0
```

packages/dart_frog/lib/src/request.dart
```diff
@@ -158,3 +154,5 @@ class Request {
     return Request._(_request.change(headers: headers, path: path, body: body));
   }
 }
+
+final _bodyFutureExpando = Expando<Future<String>>('dart_frog.request.body');
```
wolfenrain, line 158: Not that familiar with Expandos, would love to see a description on the PR to explain how this improves performance

packages/dart_frog/CHANGELOG.md
kevmoo: Fancy!!

packages/dart_frog/lib/src/request.dart
kevmoo: You all would have to measure it. 

On the VM, they're pretty cheap. And now you're avoiding copying everything all the time which should be a lot cheaper

tomarra:
Hi @kevmoo 👋  Thanks for opening this PR. Would you be able to fill in the description of the PR with more details on what the goal of this change is and how it accomplishes that? I added our normal PR template back into the description so all you should need to do is fill out the description section with those details for us to do a proper review.

kevmoo:
@tomarra – done!

kevmoo:
Do ya'll have benchmarks we cool look at?

tomarra:
> Do ya'll have benchmarks we could look at?

We do not. There are some benchmarks that are out there which you may be able to run locally, https://github.com/sharkbench/sharkbench comes to mind given their website https://sharkbench.dev/web/dart-dartfrog, to help validate this.

alestiago:
I found this [piece of documentation](https://github.com/flutter/flutter/blob/010d31d1093262fcbcb3f8672db74a713164810a/docs/contributing/Style-guide-for-Flutter-repo.md?plain=1#L851) from the Flutter Style:

> Generally speaking, `Expando` objects are a sign of an architectural problem. Carefully consider whether your usage is actually necessary. When your classes have clear documented ownership rules, there is usually a better solution.
>
> Expando objects tend to invite code that is hard to understand because one cannot simply follow references to find all the dependencies.

This said, this usage could be an exception.

kevmoo:
I think it depends on usage. I wouldn't land this PR unless there were performance tests run first!

felangel (approved):
LGTM ✅

I ran some benchmarks locally and the `Expando` refactor does have a measurable performance impact.

For example, with a request body size of 1,071,472 bytes, and 100,000,000 calls to `request.body`, using `Expando` performed approximately 3% better (~500ms faster in the test I ran).

@kevmoo I expected to see a more significant performance improvement but since there is still a measurable performance gain and we're able to make the internal `request` final again, I think it's worth merging.

felangel:
> 100,000,000

Just took a quick look and the memory impact is much more significant in the test case I ran. 

> It was a bit of an experiment on my side to see if this pattern would work.
> 
> It'd LOVE to know MEMORY impact, too. Might be interesting to look at.

Just took a quick peak and memory impact seems to be more significant. I'll post some more detailed benchmarks/snapshots in a bit but thanks again for the contribution and glad we were finally able to land it 🎉

PR #1619, opened 2024-12-11
chore(dart_frog): v1.2.0

### LINKED ISSUE #1616: chore: dart_frog release v1.2.0

tomarra:
## Status

**READY**

## Description

Closes #1616. Get the 1.2.0 release for the main dart_frog package released.

Note: Docs updates have been removed from the changelog.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore

### DIFF
--- a/packages/dart_frog/lib/src/response.dart
+++ b/packages/dart_frog/lib/src/response.dart
@@ -95,7 +95,7 @@ class Response {
           encoding: encoding,
         );
 
-  /// A shelf.Response.context key used to determine if if the
+  /// A `shelf.Response.context` key used to determine if if the
   /// [HttpResponse.bufferOutput] should be enabled or disabled.
   ///
   /// See also:
--- a/packages/dart_frog/lib/src/response.dart
+++ b/packages/dart_frog/lib/src/response.dart
@@ -95,7 +95,7 @@ class Response {
           encoding: encoding,
         );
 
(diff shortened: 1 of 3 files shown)


packages/dart_frog/lib/src/response.dart
```diff
@@ -95,7 +95,7 @@ class Response {
           encoding: encoding,
         );
 
-  /// A [shelf.Response.context] key used to determine if if the
```
tomarra, line 98: So I think this is actually a bug in the lint rule. `shelf` is included as part of an import and it doesn't seem to realize that. The simple fix is just removing the brackets so it just reads it as text instead of a object link.

packages/dart_frog/lib/src/response.dart
```diff
@@ -95,7 +95,7 @@ class Response {
...
+  /// A shelf.Response.context key used to determine if if the
```
wolfenrain, line 98:
```suggestion
  /// A `shelf.Response.context` key used to determine if if the
```

packages/dart_frog/lib/src/response.dart
tomarra: Gonna resolve this given the update below from Jochum

wolfenrain (approved):

PR #1632, opened 2024-12-18
feat(dart_frog_auth): add custom unauthenticated responses to all aut…

### LINKED ISSUE #1631: feat(dart_frog_auth): Allow users to customize their response when unauthenticated.
### Description

Currently, when not authenticated, the middleware in `dart_frog_auth` sends a response that has an empty body and status code 403. However, in applications like web sites, this is not ideal. Typically you would want to respond with an html body that informs the user they aren't logged in.

I propose adding an optional parameter to all the middleware called `unauthenticatedResponse` that takes a `FutureOr<Response> Function(RequestContext context)?`. If nothing is passed to the parameter, then it will use the existing default. If a function is passed, it will be returned if the request is not authenticated.

```dart
basicAuthentication<User>(
    unauthenticatedResponse: (context) => Response(body: /** HTML body */, status: HttpStatus.unauthenticated),
    authenticator: (context, username, password) {
        final userRepository = context.read<UserRepository>();
        return userRepository.fetchFromCredentials(username, password);
    },
)
```

### Requirements

- [ ] All CI/CD checks are passing.
- [ ] There is no drop in the test coverage percentage.
- [ ] The `unauthenticatedResponse` parameter exists on all the middleware
- [ ] The change is documented in the Dart Frog docs in the authentication page

### Additional Context

_No response_

mtwichel:
…h middleware

## Status

**READY**

## Description

Closes #1631. Adds an optional `unauthenticatedResponse` to all auth middleware which enables users to customize the response returned when a user is unauthenticated. 

## Type of Change

- [ ] 🧹 Code refactor
- [x] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/dart_frog_auth/lib/src/dart_frog_auth.dart
+++ b/packages/dart_frog_auth/lib/src/dart_frog_auth.dart
@@ -72,6 +72,7 @@ Middleware basicAuthentication<T extends Object>({
     String password,
   )? authenticator,
   Applies applies = _defaultApplies,
+  Handler? unauthenticatedResponse,
 }) {
   assert(
     userFromCredentials != null || authenticator != null,
@@ -102,7 +103,8 @@ Middleware basicAuthentication<T extends Object>({
           }
         }
 
-        return Response(statusCode: HttpStatus.unauthorized);
+        return unauthenticatedResponse?.call(context) ??
+            Response(statusCode: HttpStatus.unauthorized);
       };
 }
 
@@ -134,6 +136,7 @@ Middleware bearerAuthentication<T extends Object>({
   Future<T?> Function(String token)? userFromToken,
   Future<T?> Function(RequestContext context, String token)? authenticator,
   Applies applies = _defaultApplies,
+  Handler? unauthenticatedResponse,
 }) {
   assert(
     userFromToken != null || authenticator != null,
@@ -162,7 +165,8 @@ Middleware bearerAuthentication<T extends Object>({
           }
         }
 
-        return Response(statusCode: HttpStatus.unauthorized);
+        return unauthenticatedResponse?.call(context) ??
+            Response(statusCode: HttpStatus.unauthorized);
       };
 }
 
@@ -195,6 +199,7 @@ Middleware cookieAuthentication<T extends Object>({
     Map<String, String> cookies,
   ) authenticator,
   Applies applies = _defaultApplies,
+  Handler? unauthenticatedResponse,
 }) {
   return (handler) => (context) async {
         if (!await applies(context)) {
@@ -210,6 +215,7 @@ Middleware cookieAuthentication<T extends Object>({
           }
         }
 
-        return Response(statusCode: HttpStatus.unauthorized);
+        return unauthenticatedResponse?.call(context) ??
+            Response(statusCode: HttpStatus.unauthorized);
       };
 }
--- a/packages/dart_frog_auth/test/src/dart_frog_auth_test.dart
+++ b/packages/dart_frog_auth/test/src/dart_frog_auth_test.dart
@@ -11,6 +11,8 @@ class _MockRequestContext extends Mock implements RequestContext {}
 
 class _MockRequest extends Mock implements Request {}
 
+class _MockResponse extends Mock implements Response {}
+
 class _User {
   const _User(this.id);
   final String id;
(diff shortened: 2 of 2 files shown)


packages/dart_frog_auth/lib/src/dart_frog_auth.dart
```diff
@@ -72,6 +72,7 @@ Middleware basicAuthentication<T extends Object>({
     String password,
   )? authenticator,
   Applies applies = _defaultApplies,
+  Handler? unauthenticatedResponse,
```
alestiago, line 75: I am yet not sure if there could be a better API that the one suggested in the PR. 

I think the main issue is that the existence of `unauthenticatedResponse` makes the programmer feel there should exist a `authenticatedResponse` counterpart. Having both exist makes me question if we could merge them into a single parameter and allow the user to dictate the logic.

I've been thinking about how would it look like if instead we had an `responseOverride` parameter where the programmer gets a user and decide how to handle the authorization; but it didn't convince me. Mainly because if you would only like to override the unauthenticated behavior the programmer would be forced to define the current default `handler(context.provide(() => user));` for authenticated cases redundantly. In addition I feel that something similar could be achieved by chaining a middleware.

I have this other idea, where I think we should make a clear distinction between `authentication` and `authorization`. As of right now, the `bearerAuthentication` is doing both, authentication and authorization (despite its name being solely `bearerAuthentication`). The authorization logic is quite basic, if the authentication is a `null` user then deny access by returning an ` HttpStatus.unauthorized` response. What we could do is introduce an `authorizer` parameter (that defaults to the current behavior `user != null`). Yet, I'm not sure if we could provide a better API by leveraging the existing Dart Frog structure, for example by defining an Authorizer middleware.

---

Regarding the current API, a change I'm sure about is that if we end up going forward with the current suggested API in this PR I would:

- Rename the parameter to remove the "Response" suffix, since it it not of type `Response` but `Handler`, here are some name alternatives: 
  - `onUnauthenticated`
  - `unauthenticatedHandler`
  - `onAuthenticationFailure`

I tend to prefer `onUnauthenticated`.

mtwichel: > we should make a clear distinction between authentication and authorization 

This makes a ton of sense to me and is possibly the root of why I felt this change was needed. But this could also be solved with #1652 (or something similar) by moving the authorization logic to the handler itself. For example
```dart
Response onRequest(RequestContext context) {
  final user = context.tryRead<User>();
  if (user == null) {
    // Unauthenticated
    return UnauthenticatedResponse();
  }
  if (!hasPermission(user)) {
    // Unauthorized
    return UnauthorizedResponse();
  }

  return RealResponse();
}
```

Overall I think this is a very transparent solution and is the most flexible. But currently this won't work because if you are unauthenticated, the middleware will just return a 401 response and your handler will never actually happen.

That would be a pretty big breaking change for the auth handlers, so maybe it could be opt-in with a parameter?

```dart
bearerAuthentication<User>(
  guarded: false // when true, a default 401 response will be returned. Otherwise, no user will be provided.
)
```

Or maybe it could be a separate middleware for that behavior?

mtwichel: @felangel I still think there's a valid discussion here - basically right now, you can't use dart_frog_auth if you want routes that are optionally authenticated. Your only option if you want routes that _might_ be authenticated is your own custom middleware.

That's _fine_, but imo it's a common enough task that it should be supported in the auth package.

packages/dart_frog_auth/lib/src/dart_frog_auth.dart
```diff
@@ -195,6 +199,7 @@ Middleware cookieAuthentication<T extends Object>({
     Map<String, String> cookies,
   ) authenticator,
   Applies applies = _defaultApplies,
+  Handler? unauthenticatedResponse,
```
marcossevilla, line 202: as @alestiago said [here](https://github.com/dart-frog-dev/dart_frog/pull/1632/files#r1900701836), I prefer renaming to `onUnauthorized`
```suggestion
  Handler? onUnauthorized,
```
since we're returning an unauthorized status by default, but I'm also good with onUnauthenticated

packages/dart_frog_auth/lib/src/dart_frog_auth.dart
erickzanardo: > basically right now, you can't use dart_frog_auth if you want routes that are optionally authenticated.

@mtwichel I don't believe this is true! You can which routes are authenticated and which ones are not: https://dart-frog.dev/advanced/authentication/#filtering-routes

Or does this not work for you for some reason?

mtwichel: Ahh @erickzanardo that's an interesting approach I hadn't considered. So in the handler you'd write something like

```dart
final user = context.read<User>();

if (user is AnonymousUser) {
  return UnathenticatedResponse();
}

if (user is AppUser) {
  return AuthenticatedResponse(user.userId);
}
```

Honestly, that is way cleaner. I think we can close this, but I might open a new PR to add this use case to the docs if that's okay with yall

PR #1652, opened 2024-12-19
feat(dart_frog): add tryRead method to RequestContext

### LINKED ISSUE #1651: feat: add tryRead<T> function to simplify reading values only if they exist instead
### Description

I have a hand extension written in all my Dart Frog projects that I believe is a good candidate for the main framework.
```dart
extension on RequestContext {
  T? tryRead<T>( ) {
    try {
       return read<T>( );
    } on ProviderNotFoundException catch (_) {
       return null;
    }
  }
}
```

As you can see, all this does is attempts reading the value, and returns null if it fails. This is extremely useful in building web sites with Dart Frog as many routes are accessible for both authenticated and unauthenticated users, they just render differently. This supports a much nicer flow to access that data.

```dart
final user = context.tryRead<User>();
if (user == null) {
  print('unauthenticated');
} else {
  print('authenticated');
}
``` 

### Requirements

- [ ] All CI/CD checks are passing.
- [ ] There is no drop in the test coverage percentage.
- [ ] Change is documented in the dependency injection docs

### Additional Context

_No response_

mtwichel:
## Status

**READY**

## Description

Closes #1651. Adds a `tryRead` function to the `RequestContext` that works exactly like `read`, but instead of throwing a `StateError` when a provided value isn't found, it return `null`. This is useful for objects that may or not be provided depending on certain factors, like a `User` being provided only if they are authenticated. The change has also been reflected in the docs site.

## Type of Change

- [ ] 🧹 Code refactor
- [x] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/dart_frog/lib/src/context.dart
+++ b/packages/dart_frog/lib/src/context.dart
@@ -43,6 +43,21 @@ This can happen if $T was not provided to the request context:
     return (value as T Function())();
   }
 
+  /// Attempt to lookup an instance of [T] from the [request] context.
+  ///
+  /// Returns `null` if [T] is not available within the provided
+  /// [request] context.
+  T? tryRead<T>() {
+    try {
+      return read<T>();
+      // Explicitly catching [StateError] as it's what it throw
+      // when [read] fails
+      // ignore: avoid_catching_errors
+    } on StateError catch (_) {
+      return null;
+    }
+  }
+
   /// Get URL parameters captured by the [Router.mount].
   /// They can be accessed from inside the mounted routes.
   Map<String, String> get mountedParams {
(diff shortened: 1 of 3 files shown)


packages/dart_frog/lib/src/context.dart
```diff
@@ -43,6 +43,21 @@ This can happen if $T was not provided to the request context:
     return (value as T Function())();
   }
 
+  /// Attempt to lookup an instance of [T] from the [request] context.
+  ///
+  /// Returns `null` if [T] is not available within the provided
+  /// [request] context.
+  T? tryRead<T>() {
```
alestiago, line 50: Nit: If we end up adding `tryRead` I think we should modify the implementation to not catch the `StateError`, but instead implement `tryRead` so it does what `read` is doing right now but without the `null` check to avoid creating an additional `StateError` object. We could then base the `read` implementation on `tryRead` for code sharing.

felangel:
Closing based on the discussion in the issue 👍

PR #1818, opened 2025-08-04
feat(dart_frog): expose captured request params

### LINKED ISSUE #773: feat: Allow routing parameters in `middleware`

felangel:
## Status

**READY**

## Description

- feat(dart_frog): expose captured request params
  - closes #773
  - closes #1262

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/dart_frog/lib/src/request.dart
```diff
@@ -135,6 +135,12 @@ class Request {
     );
   }
 
+  /// Returns the url parameters captured by the [Router].
+  /// Returns an empty map if no parameters are captured.
+  ///
+  /// The returned map is unmodifiable.
```
alestiago, line 141: nit: the signature doesn't guarantee is unmodifiable and there doesn't seem to be any test in place to guarantee params is unmodifiable.

alestiago (approved):

packages/dart_frog/lib/src/request.dart
felangel: Yeah good call. I'll adjust -- also want to think a bit about whether we can consolidate this and `mountedParams` which was added a while ago.

Will make some more adjustments later today, thanks for the review!

erickzanardo (dismissed):

Tananga:
Great to see it in production!

